PR #25, opened 2023-02-13
Schema endpoint tests

YemreGurses:


tofhir-server/src/main/scala/io/tofhir/server/service/schema/SchemaFolderRepository.scala
```diff
@@ -107,7 +107,7 @@ class SchemaFolderRepository(schemaRepositoryFolderPath: String, projectFolderRe
    * @return
    */
   override def putSchema(projectId: String, id: String, schemaDefinition: SchemaDefinition): Future[Unit] = {
-    if (!schemaDefinitions.contains(projectId) || !schemaDefinitions(projectId).contains(schemaDefinition.id)) {
+    if (!schemaDefinitions.contains(projectId) || !schemaDefinitions(projectId).contains(schemaDefinition.id) || !schemaDefinitions(projectId).contains(id)) {
```
suatgonul, line 110: Why do we need: `!schemaDefinitions(projectId).contains(id)`? `id` should always be same with `schemaDefinition.id`

YemreGurses: I thought it should be better to return 404 when schema_id in URL is not found

PR #29, opened 2023-02-20
Mapping rest services

dogukan10:


tofhir-server/src/main/scala/io/tofhir/server/model/MappingMetadata.scala
```diff
@@ -0,0 +1,9 @@
+package io.tofhir.server.model
+
+/**
+ * Mapping model in the projects.json file
+ * @param id Identifier of the mapping
+ * @param url URL of the mapping
+ * @param name Name of the mapping file
+ */
+case class MappingMetadata(id: String, url: String, name: String)
```
sinaci, line 9: We already have a MappingDefinition class, but is it because we have FhirMappingExpression sequence in that class, that you needed a standalone metadata class?

tofhir-server/src/main/scala/io/tofhir/server/endpoint/MappingEndpoint.scala
```diff
@@ -11,64 +11,74 @@ import io.tofhir.server.endpoint.MappingEndpoint.SEGMENT_MAPPING
...
+      complete {
+        service.getMapping(projectId, id) map {
+          case Some(fhirMapping) => StatusCodes.OK -> fhirMapping
+          case None => StatusCodes.NotFound -> s"Mapping with name $id not found"
+        }
+      }
+    }
+  }
+
+  private def updateMapping(projectId: String, id: String): Route = {
```
sinaci, line 63: I would name this method putMapping and then the method of the service updateMapping.

tofhir-server/src/main/scala/io/tofhir/server/model/MappingMetadata.scala
sinaci: If we must create such a metadata class, I would name it FhirMappingMetadata. But first, let's check whether we can use FhirMapping by converting the non-metadata fields to Option or not.

YemreGurses: Actually, not because of just FhirMappingExpression sequence but also other fields (source, context, variable). We dont want those fields in the projects.json file. If we change those fields to Option, many in the engine codebase will be effected.

sinaci: OK. I recommend to rename it to FhirMappingMetadata and write our reasoning for its creation as a comment at the beginning of class definition.

PR #32, opened 2023-02-21
:sparkles: feature(JobService): add job services

YemreGurses:


tofhir-server/src/main/scala/io/tofhir/server/endpoint/JobEndpoint.scala
```diff
@@ -0,0 +1,89 @@
...
+          StatusCodes.NoContent
+        }
+      }
+    }
+  }
+
+}
+
+object JobEndpoint {
+  val SEGMENT_JOB = "job"
```
suatgonul, line 88: Change job into jobs.
In general, segments in REST URLs should be plural e.g. projects/<pid>/jobs or projects/<pid>/mappings/<mid>

tofhir-server/src/main/scala/io/tofhir/server/model/Project.scala
```diff
@@ -17,7 +17,7 @@ case class Project(id: String = UUID.randomUUID().toString,
                    schemas: Seq[SchemaDefinition] = Seq.empty,
                    mappings: Seq[FhirMappingMetadata] = Seq.empty,
                    contextConceptMaps: Seq[FhirMappingContext] = Seq.empty,
-                   mappingJobs: Seq[FhirMappingJob] = Seq.empty
+                   mappingJobs: Seq[JobMetadata] = Seq.empty
```
suatgonul, line 20: I think we should not use an implementation-specific model (JobMetadata) inside a general-purpose model (Project)

tofhir-server/src/main/scala/io/tofhir/server/model/JobMetadata.scala
```diff
@@ -0,0 +1,8 @@
+package io.tofhir.server.model
```
suatgonul, line 1: We might not need Metadata classes at all if we find some other way to obtain implementation-specific metadata

tofhir-server/src/main/scala/io/tofhir/server/service/job/JobRepository.scala
```diff
@@ -0,0 +1,195 @@
...
+import org.json4s.jackson.Serialization.writePretty
+
+import java.io.{File, FileWriter}
+import java.nio.charset.StandardCharsets
+import scala.collection.mutable
+import scala.concurrent.Future
+import scala.io.Source
+import io.tofhir.engine.util.FhirMappingJobFormatter.formats
+
+class JobRepository(jobRepositoryFolderPath: String, projectFolderRepository: ProjectFolderRepository) extends IJobRepository {
```
suatgonul, line 20: Rename to JobFolderRepository

PR #33, opened 2023-02-23
:sparkles: feature(MappingContext): add fhir mapping context rest ser…

YemreGurses:
…vices


tofhir-server/src/main/scala/io/tofhir/server/endpoint/MappingContextEndpoint.scala
```diff
@@ -0,0 +1,102 @@
...
+            deleteMappingContext(projectId, id)
+          } ~ pathPrefix(SEGMENT_CONTENT) {
+            pathEndOrSingleSlash {
+              uploadDownloadMappingContextRoute(projectId, id)
+            }
+          }
+        }
+    }
+  }
+
```
dogukan10, line 37: Let's add a comment for each route

tofhir-server/src/main/scala/io/tofhir/server/service/mappingcontext/MappingContextFolderRepository.scala
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Delete the mapping context from the repository
+   *
+   * @param projectId project id the mapping context belongs to
+   * @param id        mapping context id
+   * @return
+   */
+  override def deleteMappingContext(projectId: String, id: String): Future[Unit] = {
+    Future {
+      if (!mappingContextDefinitions.contains(projectId) || !mappingContextDefinitions(projectId).contains(id)) {
```
dogukan10, line 86: It is repeated three times. We can create a method to check the existence of a mapping context and use it.

tofhir-server/src/main/scala/io/tofhir/server/service/mappingcontext/MappingContextFolderRepository.scala
```diff
@@ -0,0 +1,179 @@
...
+    val map = mutable.Map.empty[String, Seq[String]]
+    val folder = FileUtils.getPath(mappingContextRepositoryFolderPath).toFile
+    if (!folder.exists()) {
+      folder.mkdirs()
+    }
+    var directories = Seq.empty[File]
+    try {
+      directories = folder.listFiles.filter(_.isDirectory).toSeq
+    } catch {
+      case e: Throwable => throw FhirMappingException(s"Given folder for the mapping context repository is not valid.", e)
```
dogukan10, line 169: Out of curiosity, do we really need try/catch block here, when does it throw an exception ?

YemreGurses: I guess you're right, we can omit that part. Also, we have same logic for other repositories (jobRepository, mappingRepository etc.), I will update those parts too.

PR #35, opened 2023-02-24
:white_check_mark: test: refactor tests, use base class for endpoint …

YemreGurses:
…tests, remove dummy folders in test-resources


tofhir-server/src/main/scala/io/tofhir/server/service/project/FolderDBInitializer.scala
```diff
@@ -36,36 +36,31 @@ class FolderDBInitializer(config: ToFhirEngineConfig,
...
+      val projects: JArray = FileOperations.readFileIntoJson(file).asInstanceOf[JArray]
+      val projectMap: Map[String, Project] = projects.arr.map(p => {
+        val project: Project = initProjectFromMetadata(p.asInstanceOf[JObject])
+        project.id -> project
+      })
+        .toMap
+      collection.mutable.Map(projectMap.toSeq: _*)
+    } else {
+      logger.debug("There does not exist a metadata file for projects. Creating it...")
+      file.getParentFile.mkdirs()
```
dogukan10, line 51: Let's make sure that parent file does not exist before creating it. Otherwise, we might delete existing content of the folder.
```suggestion
     if (!file.getParentFile.exists()) {
        file.getParentFile.mkdir()
      }
```

PR #37, opened 2023-03-06
:sparkles: feature(MappingFolderRepository): add relative context fil…

YemreGurses:
…epath handling in mappings


tofhir-server/src/main/scala/io/tofhir/server/service/mapping/MappingFolderRepository.scala
```diff
@@ -192,7 +243,7 @@ class MappingFolderRepository(mappingRepositoryFolderPath: String, projectFolder
         val source = Source.fromFile(file, StandardCharsets.UTF_8.name()) // read the JSON file
         val fileContent = try source.mkString finally source.close()
         val fhirMapping = fileContent.parseJson.extract[FhirMapping]
-        fhirMappingMap.put(fhirMapping.id, fhirMapping)
+        fhirMappingMap.put(fhirMapping.id, removeRelativePaths(fhirMapping))
```
dogukan10, line 246: Why are we removing relative path here ?

PR #38, opened 2023-03-07
:sparkles: feature(FhirMappingFolderRepository): handle $CONTEXT_REPO…

YemreGurses:
… placeholder

### DIFF
--- a/tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMapping.scala
+++ b/tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMapping.scala
@@ -82,6 +82,14 @@ object FhirMappingContextCategories {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMapping.scala
```diff
@@ -81,6 +81,10 @@ object FhirMappingContextCategories {
   final val UNIT_CONVERSION_FUNCTIONS = "unit-conversion"
 }
 
+object FhirMappingContextUrlPlaceHolder {
+  final val CONTEXT_REPO = "$CONTEXT_REPO"
```
dogukan10, line 85: We can add a simple comment explaining what is the aim of this placeholder.

dogukan10 (approved):

PR #39, opened 2023-03-09
Job running

YemreGurses:


tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
```diff
@@ -138,8 +158,48 @@ override def getJob(projectId: String, id: String): Future[Option[FhirMappingJob
...
   }
 
+  /**
+   * Run the job
+   *
+   * @param projectId project id the job belongs to
+   * @param id        job id
+   * @return
+   */
+  override def runJob(projectId: String, id: String): Future[Unit] = {
```
sinaci, line 168: There is an inconsistency in the return type of this function. FhirMappingJobManager.executeMappingJob already returns a Future and here it is wrapped in another Future; it becomes Future[Future[]]. This needs to be handled conveniently together with the streaming jobs.

PR #67, opened 2023-06-14
:bug: fix(FhirMapping): remove fields starting with @ from the mapping expression JSON

YemreGurses:
Fixes: https://gitlab.srdc.com.tr/onfhir/tofhir-web/-/issues/38

### DIFF
--- a/tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMapping.scala
+++ b/tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMapping.scala
@@ -46,6 +46,21 @@ case class FhirMapping(id: String,
       )
     )
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMapping.scala
```diff
@@ -46,6 +46,21 @@ case class FhirMapping(id: String,
       )
     )
   }
+
+  /**
+   * Removes any fields starting with @ from the mapping expression JSON.
+   * @return
+   */
+  def removeAtFields(): FhirMapping = {
```
dogukan10, line 54: Could you please elaborate on the usage/meaning of these fields i.e. what are they ?

I think we should rename the function name as well if possible.

PR #72, opened 2023-07-06
Fix #71

### LINKED ISSUE #71: Same error logs should be generated when `mappingErrorHandling` paramater of a job is set `continue` or `halt`
When this setting is set to `halt`, an error log is generated for a resource that cannot be mapped. However, there is no such error log in case of `continue`

dogukan10:


tofhir-engine/src/main/scala/io/tofhir/engine/data/write/SinkHandler.scala
```diff
@@ -40,44 +42,29 @@ object SinkHandler {
...
-        )
+      logMappingJobResult(mappingJobExecution,mappingUrl,numOfFhirResources,fhirWriteProblemsAccum.value,mappingErrors,invalidInputs)
       //Unpersist the data frame
       df.unpersist()
     } catch {
       case t:Throwable => {
         t.getCause match {
           // FhirMappingInvalidResourceException is already handled above
           case e:FhirMappingInvalidResourceException => None
+          // FhirMappingException is already handled and logged by Spark while running the mapping
```
sinaci, line 58: What do you mean by "already handled and logged by Spark"?

tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingTaskExecutionException.scala
```diff
@@ -0,0 +1,10 @@
+package io.tofhir.engine.model
+
+/**
+ * Exception to be thrown when there is an error during mapping task execution.
+ *
+ * @param msg the message
+ * */
+final case class FhirMappingTaskExecutionException(private val msg: String)
```
sinaci, line 8: Up to now we have been using the FhirMappingException class to throw and handle the expcetions throughout the system. We recently added FhirMappingInvalidResourceException class to specifically identify and handle the errors occuring while communicating with the FHIR endpoint. Now we are adding FhirMappingTaskExecutionException to differentiate another specific error during the task execution.

Shall we have the FhirMappingException as the umbrella exception class for the whole system and the others extending this for specific things? Then shall we walk through all places that we throw/handle FhirMappingException to check whether we use extended classes there or not? Maybe we can discuss these in a meeting.

sinaci (changes requested):

tofhir-engine/src/main/scala/io/tofhir/engine/data/write/SinkHandler.scala
dogukan10: When FhirMappingException is thrown, Spark will encapsulate it with SparkException and prints its log. Therefore, we do not need to handle this exception here, otherwise we end up with duplicate logs. Therefore, I simply skipped these exceptions which are already logged by Spark.

tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingTaskExecutionException.scala
dogukan10: I've just removed that class and used FhirMappingException instead 👍

PR #74, opened 2023-07-10
Update job terminology

Okanmercan99:


tofhir-server/src/main/scala/io/tofhir/server/endpoint/CodeSystemEndpoint.scala
```diff
@@ -48,22 +48,22 @@ class CodeSystemEndpoint extends LazyLogging {
...
-          service.createCodeSystem(terminologyId, codeSystem) map { created =>
-            StatusCodes.Created -> created
-          }
-        }
-      }
-    }
-  }
+//  /**
+//   * Route to create a code system within a terminology
+//   * TODO: Implement job terminology update feature before using this
```
dogukan10, line 53: What does that mean ? I think that feature is already implemented in this MR ?

tofhir-server/src/main/scala/io/tofhir/server/service/TerminologySystemService.scala
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Update the TerminologySystem
+   * @param id TerminologySystem id
+   * @param terminologySystem TerminologySystem to update
+   * @return Updated TerminologySystem
+   */
+  def updateTerminologySystem(id: String, terminologySystem: TerminologySystem): Future[TerminologySystem] = {
+
+    // Get project ids from project repository
+    val projectIds: Future[Seq[String]] = projectRepository.getAllProjects.map(_.map(_.id))
```
dogukan10, line 51: I think that we do not need to fetch project ids. We can simply make use of **mappingJobRepository.getCachedMappingsJobs** method which also includes project id. Thanks to this change, we can remove projectRepository from the class parameters.

tofhir-server/src/main/scala/io/tofhir/server/service/TerminologySystemService.scala
```diff
@@ -0,0 +1,134 @@
...
+    projectIds.flatMap { projectIds =>
+        Future.traverse(projectIds) { projectId =>
+
+        // Get jobs of given project
+        val jobs: Future[Seq[FhirMappingJob]] = mappingJobRepository.getAllJobs(projectId)
+
+        jobs.map { jobs =>
+          jobs
+            .filter { job =>
+              job.terminologyServiceSettings.isDefined && job.terminologyServiceSettings.get.isInstanceOf[LocalFhirTerminologyServiceSettings]
```
dogukan10, line 62: This filter can also include the if check we used in line 70. Further, that part is duplicated in line 111,  therefore you can implement a simple method to be passed into the filter method here.

tofhir-server/src/main/scala/io/tofhir/server/service/TerminologySystemService.scala
```diff
@@ -0,0 +1,134 @@
...
+    // Update TerminologySystem
+    terminologySystemRepository.updateTerminologySystem(id, terminologySystem)
+  }
+
+  /**
+   * Delete the TerminologySystem
+   * @param id TerminologySystem id
+   * @return
+   */
+  def deleteTerminologySystem(id: String): Future[Unit] = {
```
dogukan10, line 97: What happens if I try to delete a non-existent terminology service ? We should handle that case.
Moreover, instead of updating the mapping job, we can give a warning indicating that this terminology service is used by some mapping jobs so you can not delete it ? WDYT @sinaci @suatgonul @YemreGurses ?

tofhir-server/src/test/scala/io/tofhir/server/terminology/TerminologyServiceManagerEndpointTest.scala
```diff
@@ -26,6 +28,11 @@ class TerminologyServiceManagerEndpointTest extends BaseEndpointTest {
 
   val terminologySystem2: TerminologySystem = TerminologySystem(name = "testTerminology2", description = "example terminology 2", codeSystems = Seq.empty, conceptMaps = Seq.empty)
 
+  // Create job for test update on job terminology
+  val sinkSettings: FhirSinkSettings = FileSystemSinkSettings(path = "http://example.com/fhir")
+  val terminologyServiceSettings: TerminologyServiceSettings = LocalFhirTerminologyServiceSettings(s"/${TerminologySystemFolderRepository.TERMINOLOGY_SYSTEMS_FOLDER}/${terminologySystem1.id}", codeSystemFiles = Seq.empty, conceptMapFiles = Seq.empty)
+  val job3: FhirMappingJob = FhirMappingJob(name = Some("mappingJob3"), sourceSettings = Map.empty, sinkSettings = sinkSettings, terminologyServiceSettings = Some(terminologyServiceSettings), mappings = Seq.empty, mappingErrorHandling = ErrorHandlingType.CONTINUE)
```
dogukan10, line 34: Rename it to job because there is no job,job1 or job2

tofhir-server/src/main/scala/io/tofhir/server/service/TerminologySystemService.scala
YemreGurses: Agress, we should check if that ts is used in a job and give relevant response.

tofhir-server/src/main/scala/io/tofhir/server/endpoint/CodeSystemEndpoint.scala
YemreGurses: I don't think this much commented code is useful. We can simply remove them instead of merging them like this. We can retrieve it from git history if needed in the future.

tofhir-server/src/main/scala/io/tofhir/server/service/TerminologySystemService.scala
```diff
@@ -0,0 +1,137 @@
...
+              }
+              // Delete terminology service case
+              case None => job.copy(terminologyServiceSettings = None)
+            }
+            // Update job in the repository
+            mappingJobRepository.putJob(projectId, job.id, updatedJob)
+          }
+      }
+    }
+    Future.apply(None)
```
YemreGurses, line 135: Why is this here? Instead, you can wrap whole method inside a Future { }

sinaci: A new Future is created for each projectId within foreach loop. Those Futures should be handled appropriately.

PR #79, opened 2023-07-18
Execution intermediate errors

Okanmercan99:


tofhir-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -145,12 +147,51 @@ class ExecutionService(jobRepository: IJobRepository, mappingRepository: IMappin
         Seq.empty
       }
       else {
-        val jobRuns = dataFrame.filter(s"executionId = '$executionId'")
-        // handle the case where the job has not been run yet which makes the data frame empty
-        if (jobRuns.isEmpty) {
+        // Get job run logs for the given execution
+        val jobRunLogs = dataFrame.filter(s"executionId = '$executionId' and projectId is not null")
```
dogukan10, line 151: Comment on why we need **projectId is not null** filter.

tofhir-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -145,12 +147,51 @@ class ExecutionService(jobRepository: IJobRepository, mappingRepository: IMappin
...
+
+        // Handle the case where the job has not been run yet, which makes the data frame empty
+        if (jobRunLogs.isEmpty) {
           Seq.empty
         } else {
-          jobRuns.collect().map(row => {
+          // Get error logs for the given execution and select needed columns
+          val jobErrorLogs = dataFrame.filter(s"executionId = '$executionId' and projectId is null").select(List("errorCode", "errorDesc", "message", "mappingUrl").map(col):_*)
```
dogukan10, line 158: Comment on why we need **projectId is null** filter. As far as I understand, these are the error logs for erroneous sources i.e. rows. If so, we could rename the variable accordingly

tofhir-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -145,12 +147,51 @@ class ExecutionService(jobRepository: IJobRepository, mappingRepository: IMappin
...
+
+          // Group job error logs by mapping url
+          val jobErrorLogsGroupedByMappingUrl = jobErrorLogs.groupByKey(row => row.get(row.fieldIndex("mappingUrl")).toString)(Encoders.STRING)
+
+          // Define a new schema for the resulting rows. We will add a "error_logs" column to job run logs that contains related error logs.
+          val updatedSchema = jobRunLogs.schema.add("error_logs", ArrayType(
```
dogukan10, line 164: Let's remove this variable and directly create it in line 172. Move the comment there.

tofhir-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -145,12 +147,51 @@ class ExecutionService(jobRepository: IJobRepository, mappingRepository: IMappin
...
+              .add("mappingUrl", StringType)
+          ))
+          // Define a row encoder for new schema
+          implicit val rowEncoder: Encoder[Row] = RowEncoder(updatedSchema)
+
+          // Collect job run logs for matching with mappingUrl field of job error logs
+          var jobRunLogsData = jobRunLogs.collect()
+
+          // Get updated job run logs
+          val updatedJobRunLogs = jobErrorLogsGroupedByMappingUrl.mapGroups((key, values) => {
```
dogukan10, line 178: Rename it to jobRunLogsWithInputErrors or jobRunLogsWithErrorDetails or similar.

Further, in the comment, let's explain what  it is being updated. Maybe, you can get rid of rowEncoder variable and create it in line 183 and combine the comments properly ?

dogukan10:
In the logs of FHIRMappingJobResult, we have the following:

```
2023-07-19_16:01:37.917 [scala-execution-context-global-33] INFO  i.t.e.d.w.SinkHandler$ - toFHIR batch mapping result (PARTIAL_SUCCESS) for execution '80b237fa-d2b4-4663-b949-4c7f03bf4510' of job 'pilot1' in project 'pilot1' for mapping 'https://aiccelerate.eu/fhir/mappings/pilot1/patient-mapping'!
	# of Invalid Rows: 	1
	# of Not Mapped: 	0
	# of FHIR resources:	9
	# of Failed writes:	0
```

I suggest reordering them as follows:

-  "# of Invalid Rows"
-  "# of Not Mapped"
-  "# of Failed writes"
-  "# of FHIR resources"

Morever, it would be good if we rename **# of FHIR resources** to **# of Written FHIR resources**.

WDYT @sinaci @suatgonul @YemreGurses ?

sinaci:
I agree with @dogukan10.

dogukan10:
While testing this branch, I noticed that we lost the ordering of selected mappings in mapping job execution some time ago.

@Okanmercan99  could you please update https://github.com/srdc/tofhir/blob/main/tofhir-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala#L65 to the following which solves this bug ?

`case Some(urls) => urls.flatMap(url => mappingJob.mappings.find(p => p.mappingRef.contentEquals(url)))`

dogukan10 (approved):

suatgonul (approved):

PR #86, opened 2023-08-08
:pencil2: rename InferTask properties

YemreGurses:


tofhir-server/src/main/scala/io/tofhir/server/model/InferTask.scala
```diff
@@ -1,11 +1,11 @@
 package io.tofhir.server.model
 
-import io.tofhir.engine.model.{DataSourceSettings, SqlSource}
+import io.tofhir.engine.model.{DataSourceSettings, FhirMappingSourceContext}
 
 /**
  * Infer task instance for getting connection settings and sql query. InferTask object is implemented only for SQL sources.
```
dogukan10, line 6: Let's fix the comment. In addition to SQL sources, it supports file system as well.

tofhir-server/src/main/scala/io/tofhir/server/model/InferTask.scala
```diff
@@ -1,11 +1,11 @@
...
  *
- * @param sourceSettings    Connection details for data source
- * @param sqlSource         Sql query for data source
+ * @param sourceSettings Connection details for data source
+ * @param fhirMappingSourceContext mapping task source information
```
dogukan10, line 9: I think that we can rename fhirMappingSourceContext to sourceContext and fix the comment. WDYT ?

PR #91, opened 2023-08-25
:sparkles: (Server): Added RxNorm and DT4H-related function libraries…

suatgonul:
… to the engine instantiated for the server

### DIFF
--- a/tofhir-server/pom.xml
+++ b/tofhir-server/pom.xml
@@ -169,11 +169,6 @@
             <groupId>io.onfhir</groupId>
             <artifactId>tofhir-rxnorm_2.13</artifactId>
         </dependency>
-        <dependency>
-            <!-- TODO get rid of this dependency when the server supports injecting libraries in some other way -->
-            <groupId>io.onfhir</groupId>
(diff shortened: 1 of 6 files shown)


tofhir-server/pom.xml
```diff
@@ -165,6 +165,15 @@
             <groupId>io.onfhir</groupId>
             <artifactId>tofhir-common</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.onfhir</groupId>
+            <artifactId>tofhir-rxnorm_2.13</artifactId>
+        </dependency>
+        <dependency>
+            <!-- TODO get rid of this dependency when the server supports injecting libraries in some other way -->
```
sinaci, line 173: This is OK for now, as you suggested; we might implement external library injection in a later release.

On the other hand, I suppose these are very specific functions for the requirements of the DT4H project. Even in this case, it might be an option to implement a general function library including these functions. What do you think? @suatgonul

suatgonul: Do you mean having a library inside the toFHIR code base instead of adding it as a dependency?

PR #96, opened 2023-09-04
Initial version of ExecutionManager

suatgonul:
- Adapted checkpointing per job and per mapping task in the job
- Implemented an initial version of the ExecutionManager that is able to manage streaming mapping tasks
- Implemented REST services to stop running streaming mapping tasks
- Streaming tasks can be run, listed and stopped in a non-blocking manner via interactive CLI. 
- When a streaming job is run directly via the CLI, all streaming queries are waited for termination.


tofhir-engine/src/main/scala/io/tofhir/engine/cli/CommandFactory.scala
```diff
@@ -10,6 +10,8 @@ object CommandFactory {
       case "extract-redcap-schemas" => new ExtractRedCapSchemas()
       case "reload" => new Reload()
       case "run" | "execute" => new Run()
+      case "list" => new ListRunningMappings()
```
dogukan10, line 13: When I type **help** (the command we use to list available commands), there is no l**ist** command

tofhir-engine/src/main/scala/io/tofhir/engine/cli/CommandFactory.scala
```diff
@@ -10,6 +10,8 @@ object CommandFactory {
...
+      case "stop" => new Stop()
```
dogukan10, line 14: In the help menu, we have the following documentation for stop command:
**stop - Stop the execution of the MappingJob (if any).**

I think that we should update it because as far as I understand, we can stop individual mappings.

dogukan10:
Please update README to include the documentations of **list** and **stop** (it includes the documentation for stop command but it can be improved) commands.

tofhir-engine/pom.xml
```diff
@@ -236,6 +236,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-scala_2.13</artifactId>
+            <scope>test</scope>
```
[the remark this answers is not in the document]
suatgonul: It seems the best practice is not to set scope in dependency management at all. So, I am keeping it here for the moment. Even, we could remove scopes from the dependency management in the future.
 
https://stackoverflow.com/questions/15221299/dependencymanagement-and-scope

PR #113, opened 2023-10-30
:bug: fix(KafkaSourceSettings): update asStream value in KafkaSourceS…

YemreGurses:
3 changes:
- Apply archiving to only file system sources
- Update KafkaSourceSettings to include asStream field in the responses from API.
- Show apache spark logs if an error or warn level log is encountered.


tofhir-engine/src/main/scala/io/tofhir/engine/util/KafkaSourceSettingsSerializers.scala
suatgonul, line 1: Can you add a simple test ensuring that we are returning as stream as expected? Updating existing job tests could be an option.

PR #115, opened 2023-10-31
Erroneous records test

YemreGurses:
add erroneous records tests after mapping execution


suatgonul:
Can you please add tests showing that erroneous records are accumulated in the corresponding file for streaming jobs as new batch, with erroneous records, are processed.

suatgonul (approved):

PR #117, opened 2023-11-02
Avoid deletion a mapping if referenced in any job 

YemreGurses:
- Avoid deletion if a mapping is referenced by any job 
- Update mapping task url in referenced jobs if any during update at mapping url field

### DIFF
--- a/tofhir-server/src/main/scala/io/tofhir/server/service/job/IJobRepository.scala
+++ b/tofhir-server/src/main/scala/io/tofhir/server/service/job/IJobRepository.scala
@@ -52,4 +52,12 @@ trait IJobRepository {
    * @return
    */
   def deleteJob(projectId: String, id: String): Future[Unit]
+
+  /**
+   * Retrieves the jobs referencing the given mapping in their definitions.
+   * @param projectId identifier of project whose jobs will be checked
+   * @param mappingUrl the url of mapping
+   * @return the jobs referencing the given mapping in their definitions
+   */
+  def getJobsReferencingMapping(projectId: String, mappingUrl: String): Future[Seq[FhirMappingJob]]
 }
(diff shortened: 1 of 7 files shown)


tofhir-server/src/main/scala/io/tofhir/server/service/job/IJobRepository.scala
```diff
@@ -52,4 +52,12 @@ trait IJobRepository {
...
    */
   def deleteJob(projectId: String, id: String): Future[Unit]
+
+  /**
+   * Retrieves the jobs referencing the given mapping in their definitions.
+   * @param projectId identifier of project whose jobs will be checked
+   * @param mappingUrl the url of mapping
+   * @return the jobs referencing the given mapping in their definitions
+   */
+  def getJobsReferencingMapping(projectId: String, mappingUrl: String): Future[Seq[FhirMappingJob]]
```
suatgonul, line 62: I am not sure whether these methods should be part of the API. We would not have needed it if had a relational db based implementation. But, let's keep them for now

tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
```diff
@@ -136,9 +136,23 @@ override def getJob(projectId: String, id: String): Future[Option[FhirMappingJob
...
+   * Retrieves the jobs referencing the given mapping in their definitions.
+   *
+   * @param projectId  identifier of project whose jobs will be checked
+   * @param mappingUrl the url of mapping
+   * @return the jobs referencing the given mapping in their definitions
+   */
+  override def getJobsReferencingMapping(projectId: String, mappingUrl: String): Future[Seq[FhirMappingJob]] = {
+    Future {
+      val jobs: Seq[FhirMappingJob] = jobDefinitions.getOrElse(key = projectId, default = Map.empty).values.toSeq
+      jobs.filter(job => job.mappings.map(mappingTask => mappingTask.mappingRef).contains(mappingUrl))
```
suatgonul, line 149: Why do we use contains here? Should not we use equals or contentEquals?

YemreGurses: This method returns job list referencing specific mapping. That's why, we filter out those that do not contain the mapping url.

PR #119, opened 2023-11-07
fix simplifier in SimpleStructureDefinition for coverChild mappings

YemreGurses:
:bug: fix: simplifier in SimpleStructureDefinition stops recursion if the profile is already visited twice now, it was one before. And since this increases json's size, an exception about Identifier is added


tofhir-server/src/main/scala/io/tofhir/server/service/SimpleStructureDefinitionService.scala
```diff
@@ -74,7 +74,12 @@ class SimpleStructureDefinitionService(fhirConfig: BaseFhirConfig) {
       }
 
       // Start of the simplifier method
-      if (profileUrl.isDefined && accumulatingTypeUrls.contains(profileUrl.get)) {
+      if (profileUrl.isDefined &&
+        // Stop the recursion if the profile is already visited twice (to avoid infinite recursion)
+        (accumulatingTypeUrls.count(_ == profileUrl.get) >= 2 ||
```
sinaci, line 79: I would implement an elimination to this so that it does not count (twice) every time. I will update this part.

PR #126, opened 2023-11-22
add test for streaming job execution

YemreGurses:


tofhir-server/src/test/scala/io/tofhir/server/project/MappingExecutionEndpointTest.scala
dogukan10, line 1: After this test is executed, it produces a checkpoint folder as you can see in the following image: 
image
Could you please modify it to delete this folder ?

dogukan10 (changes requested):

dogukan10 (approved):

PR #128, opened 2023-11-29
Execution table filter

camemre49:
- Filter functionality is added according to the start time and error status columns of the executions table of a selected job. 
- New test file added to the log server.
- Filter information is sent as query parameters.
- A bug related to mapping test result is fixed: https://github.com/srdc/tofhir/pull/128/commits/48c275d970399ff68e8b864a6b92a928bb25a0af


dogukan10:
In ExecutionService, we have a hardcoded value i.e. **logs/tofhir-mappings.log** of the path to the log file. I think we should read it from the application.conf.

tofhir-log-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -150,17 +162,32 @@ class ExecutionService() extends LazyLogging {
...
+            if(Option(dateAfter).nonEmpty){
+              filteredLogs = filteredLogs.filter(col("startTime") > dateAfter)
+            }
+            if (Option(dateBefore).nonEmpty) {
+              filteredLogs = filteredLogs.filter(col("startTime") < dateBefore)
+            }
+
             // sort the executions by latest to oldest
-            val paginatedLogs = executionLogs.sort(executionLogs.col("startTime").desc).collect().slice(start, end)
+            val paginatedLogs = filteredLogs.sort(executionLogs.col("startTime").desc).collect().slice(start, end)
```
dogukan10, line 185: **start** and **end** parameters are calculated based on the **total** variable which is wrong because we need to find these variables after the filtering is done. 

Further, please update the tests to reproduce this error. To do this, you can set page size to 4 and simply pass "page" -> "2" parameter to "should get executions filtered by a date range" test. 

In addition to this bug, I think it would be great if you can update the function so that it takes the page size as a query parameter whose default value is 10 which is hardcoded now.

dogukan10 (changes requested):

tofhir-server/src/main/scala/io/tofhir/server/interceptor/ICORSHandler.scala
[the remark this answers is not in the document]
dogukan10: Please ignore this comment, this work is under development in another branch. Resolving the thread.

tofhir-server/src/main/scala/io/tofhir/server/util/LogServiceClient.scala
```diff
@@ -28,24 +28,27 @@ class LogServiceClient(logServiceEndpoint: String) {
    * @param projectId
    * @param jobId
    * @param page
+   * @param filters start end, end date, error status and running status information to filter executions. Uri is formatted in frontend.
```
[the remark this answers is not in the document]
camemre49: Here, format is a string like errorStatus=SUCCESS&dateAfter=2020-01-01. So they are sent to the log-server as different query parameters.
- It is created in front-end and passed as a query parameter called 'filters'.

tofhir-log-server/src/test/scala/io/tofhir/server/service/ExecutionServiceTest.scala
```diff
@@ -0,0 +1,125 @@
...
+          (logs.head \ "mappingUrl").extract[String] shouldEqual "https://aiccelerate.eu/fhir/mappings/pilot1/medication-administration-mapping"
+          (logs(1) \ "result").extract[String] shouldEqual "SUCCESS"
+          (logs(1) \ "numOfFhirResources").extract[Int] shouldEqual 5
+          (logs(1) \ "mappingUrl").extract[String] shouldEqual "https://aiccelerate.eu/fhir/mappings/pilot1/medication-administration-mapping"
+
+        })
+    }
+  }
+
+  override def beforeAll(): Unit = {
```
dogukan10, line 112: Add an application.conf file for test environment and configure the tofhir.logs-path accordingly. Then, remove this redundant beforeAll function.

tofhir-server/src/main/scala/io/tofhir/server/util/LogServiceClient.scala
```diff
@@ -27,15 +27,26 @@ class LogServiceClient(logServiceEndpoint: String) {
    *
    * @param projectId
    * @param jobId
-   * @param page
-   * @return A future of a tuple containing the details of individual executions and total number of executions
+   * @param page desired number of page
+   * @param rowPerPage row count per page
+   * @param dateBefore last date of the filtered executions
+   * @param dateAfter start date of the filtered executions
+   * @param errorStatuses desired error status of the filtered executions
```
dogukan10, line 34: This param includes a list of error status separated by comma. Update the documentation, accordingly.

tofhir-log-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -29,15 +32,22 @@ class ExecutionService() extends LazyLogging {
...
         Seq.empty
       }
       else {
         // Get mapping tasks logs for the given execution. ProjectId field is not null for selecting mappingTasksLogs, filter out row error logs.
-        val mappingTasksLogs = dataFrame.filter(s"executionId = '$executionId' and projectId is not null")
+        var mappingTasksLogs = dataFrame.filter(s"executionId = '$executionId' and projectId is not null")
+
+        // If the level of the mapping task log is 'ERROR', mark the log as 'FAILURE'
+        // Mapping tasks without input data stay marked as 'STARTED' but its level field is 'ERROR', map it to 'FAILURE'.
+        mappingTasksLogs = mappingTasksLogs.withColumn("result",
```
dogukan10, line 47: Although I understand why we need this code here, it seems that there is a bug which needs to be solved in the logging. For example, in the test log file, I see that we get a "FHIR repository at url http://localhost:8081/fhir returned an unidentified error while writing the resources" exception, and its result is STARTED and we have 
```
"numOfInvalids": -1,
"numOfNotMapped": -1,
"numOfFhirResources": -1,
"numOfFailedWrites": -1
```
which does not make sense because we get the error while writing the resources so numOfFailedWrites should be set to something.

tofhir-log-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -144,23 +155,45 @@ class ExecutionService() extends LazyLogging {
               StructField("startTime", StringType) ::
               StructField("errorStatus", StringType) :: Nil
           )))
-          // page size is 10, handle pagination
+          val pageSize: Int = queryParams.getOrElse("rowsPerPage", "10").toInt
+          // handle the pagination according to the page size
           val total = executionLogs.count()
```
dogukan10, line 160: This variable does not make sense any more becase we have a filtering. Please modify the code so that start and end variables are calculated based on the filtered logs correctly.

dogukan10 (changes requested):

tofhir-log-server/src/test/scala/io/tofhir/server/service/ExecutionServiceTest.scala
camemre49: I do not understand, why the function is redundant and how it may be excluded by using a conf file.

tofhir-log-server/src/main/scala/io/tofhir/server/config/ToFhirConfig.scala
```diff
@@ -0,0 +1,25 @@
+package io.tofhir.server.config
+
+
+import com.typesafe.config.Config
+
+/**
+ * ToFhir log-server configurations
+ */
+object ToFhirConfig {
```
dogukan10, line 9: Rename it as ToFhirLogServerConfig and move it to io.tofhir.log.server package

dogukan10 (changes requested):

dogukan10 (approved):

sinaci (changes requested):

PR #129, opened 2023-11-29
:sparkles: feat: Start versioning toFHIR from 1.0.0. The Maven versio…

sinaci:
…n will be logged at the very beginning of the application as app version.


docker/engine/build.sh
```diff
@@ -1,4 +1,5 @@
 # Execute one of the following commands from the project.root.directory (../../)
+# Do not forget to use the correct version while tagging.
```
dogukan10, line 2: Actually, we can utilize the Maven Versions Plugin and Docker Maven Plugin to create the Docker image with correct tag. WDYT ?

sinaci: Docker Maven Plugin seems like a good utility. But, if I understand correctly, we will need to heavily configure it so that it will use the docker-entrypoint.sh file. On the other hand we have multiple images which are created and used in different deployment settings. For now, I am in favor of following this manual approach.

PR #138, opened 2023-12-11
Handle the case where there is no project directories provided in the configured folders

dogukan10:
Each repository directory should include project directories. For example, for the mapping repository, we should have:

- mappings
  - project1
    - project1-mapping.json
  - project2
    - project2.-mapping.json
  
If there are some files instead of directories inside mappings folder, the initialization of tofhir-server fails. This MR handles such problems.

### DIFF
--- a/tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
+++ b/tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
@@ -205,10 +205,6 @@ override def getJob(projectId: String, id: String): Future[Option[FhirMappingJob
     }
     var directories = Seq.empty[File]
     directories = folder.listFiles.filter(_.isDirectory).toSeq
-    if (directories.isEmpty) {
-      throw new RuntimeException("There is no project folder for job in the repository")
-    }
-
     directories.foreach { projectDirectory =>
       // job-id -> FhirMappingJob
       val fhirJobMap: mutable.Map[String, FhirMappingJob] = mutable.Map.empty
--- a/tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
+++ b/tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
@@ -1,5 +1,6 @@
 package io.tofhir.server.service.job
 
+import com.fasterxml.jackson.core.JsonParseException
 import com.typesafe.scalalogging.Logger
 import io.onfhir.api.util.IOUtil
 import io.onfhir.util.JsonFormatter._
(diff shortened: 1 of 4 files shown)


tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
```diff
@@ -191,6 +191,10 @@ override def getJob(projectId: String, id: String): Future[Option[FhirMappingJob
     }
     var directories = Seq.empty[File]
     directories = folder.listFiles.filter(_.isDirectory).toSeq
+    if (directories.isEmpty) {
+      throw new RuntimeException("There is no project folder for job in the repository")
```
dogukan10, line 195: Let's remove this commit because it prevents tofhir-server from starting when there is no project. We'll implement the correct logic in a new commit.

tofhir-server/src/main/scala/io/tofhir/server/service/mapping/ProjectMappingFolderRepository.scala
```diff
@@ -229,10 +234,17 @@ class ProjectMappingFolderRepository(mappingRepositoryFolderPath: String, projec
...
-          fhirMappingMap.put(fhirMapping.id, fhirMapping)
+        // Try to parse the file content as FhirMapping
+        try {
+          val fhirMapping = fileContent.parseJson.extract[FhirMapping]
+          // discard if the mapping id and file name not match
+          if (FileOperations.checkFileNameMatchesEntityId(fhirMapping.id, file, "mapping")) {
+            fhirMappingMap.put(fhirMapping.id, fhirMapping)
+          }
+        }catch{
+          case _: JsonParseException =>
```
dogukan10, line 245: How can I reproduce this error ?

I utilize an empty json file and it gives me a org.json4s.MappingException exception. Let's catch all exception with case e:Throwable and log the original error as well because it provides the real problem in the json like "No usable value for id"

tofhir-server/src/main/scala/io/tofhir/server/service/job/JobFolderRepository.scala
```diff
@@ -212,10 +213,17 @@ override def getJob(projectId: String, id: String): Future[Option[FhirMappingJob
...
-          fhirJobMap.put(job.id, job)
+        // Try to parse the file content as FhirMappingJob
+        try{
+          val job = fileContent.parseJson.extract[FhirMappingJob]
+          // discard if the job id and file name not match
+          if(FileOperations.checkFileNameMatchesEntityId(job.id, file, "job")) {
+            fhirJobMap.put(job.id, job)
+          }
+        }catch{
+          case _: JsonParseException =>
```
dogukan10, line 224: Please see my comment about the mapping.

dogukan10 (commented):
It's OK, we can merge it after the build passes.

PR #140, opened 2023-12-12
Log server refactor

Okanmercan99:
- Project endpoint is removed from tofhir-log-server.
- Job endpoint is removed from tofhir-log-server.
- ToFhirServerEndpoint is renamed as ExecutionEndpoint in tofhir-log-server.
- ToFhirServer is renamed as ToFhirLogServer in tofhir-log-server.


dogukan10 (approved):

tofhir-log-server/src/main/scala/io/tofhir/log/server/ToFhirHttpServer.scala
```diff
@@ -1,11 +1,11 @@
-package io.tofhir.server
+package io.tofhir.log.server
```
sinaci, line 1: I suggest to rename ToFhirHttpServer also to sth like ToFhirLogHttpServer

tofhir-log-server/src/main/scala/io/tofhir/log/server/endpoint/ExecutionEndpoint.scala
```diff
@@ -1,41 +1,46 @@
...
-class JobEndpoint() extends LazyLogging {
+class ExecutionEndpoint(webServerConfig: WebServerConfig) extends ICORSHandler with IErrorHandler with LazyLogging {
 
   val executionService: ExecutionService = new ExecutionService()
 
-  def route(request: ToFhirRestCall): Route = {
-    pathPrefix(SEGMENT_JOB) {
-      val projectId: String = request.projectId.get
-      pathPrefix(Segment) { jobId: String =>
-        pathPrefix(SEGMENT_EXECUTIONS) { // Operations on all executions, jobs/<jobId>/executions
```
sinaci, line 23: There are 2 comments like this -> // Operations on all executions, jobs/<jobId>/executions

Are there any operations on these endpoint? If not, let's update the comments.

sinaci (changes requested):

tofhir-log-server/src/main/scala/io/tofhir/log/server/ToFhirHttpServer.scala
Okanmercan99: I intended to move this class to 'tofhir-server-common' since it's identical to the one in 'tofhir-server.'

PR #142, opened 2023-12-14
:recycle: refactor: Remove error handling from FhirMappingJobManager …

Okanmercan99:
…and MappingTaskExecutor.


tofhir-engine/src/main/scala/io/tofhir/engine/data/write/FhirRepositoryWriter.scala
```diff
@@ -145,76 +145,60 @@ class FhirRepositoryWriter(sinkSettings: FhirRepositorySinkSettings, runningJobR
...
-                code = FhirMappingErrorCodes.FHIR_API_TIMEOUT,
+                code = FhirMappingErrorCodes.SERVICE_PROBLEM,
                 description = msg
               )))
             ).foreach(failedResult => problemsAccumulator.add(failedResult))
           None
         }
+      case tout: TimeoutException =>
+        val msg = s"FHIR repository at url ${sinkSettings.fhirRepoUrl} timeout for batch interaction while writing the resources!"
+        logger.error(msg, tout)
```
dogukan10, line 180: We did not have this logging in the previous version. This may lead duplicate logs since we pass the error to problemsAccumulator ?

tofhir-engine/src/main/scala/io/tofhir/engine/mapping/FhirMappingJobManager.scala
```diff
@@ -84,17 +81,10 @@ class FhirMappingJobManager(
               val jobResult = FhirMappingJobResult(mappingJobExecution, Some(task.mappingRef))
               logger.error(jobResult.toLogstashMarker, jobResult.toString, se)
           }
-        // Halt the execution if error handling type is HALT
-        haltOrContinueExecution(mappingJobExecution, task, se)
-        // Pass the stop exception to the upstream Futures in the chain laid out by foldLeft above
-        case t: FhirMappingJobStoppedException =>
```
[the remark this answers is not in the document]
Okanmercan99: Since it is a foldLeft chain, When we throw it, it interrupts the next mappingTask, and the function goes to recover part again and again. When we remove this throw t, it is caught in case e: Throwable, and It does not interrupt other mappingTasks.

tofhir-engine/src/main/scala/io/tofhir/engine/data/write/FhirRepositoryWriter.scala
```diff
@@ -144,77 +144,57 @@ class FhirRepositoryWriter(sinkSettings: FhirRepositorySinkSettings, runningJobR
...
-            .map(_._2)
-            .map(mr =>
-              mr.copy(error = Some(FhirMappingError( //Set the error
-                code = FhirMappingErrorCodes.SERVICE_PROBLEM,
-                description = msg + " " + e.getMessage
-              )))
-            ).foreach(failedResult => problemsAccumulator.add(failedResult))
-          None
-        }
+        logger.error(msg, e)
```
dogukan10, line 188: Let's remove this logging which does not exist in the previous version.

tofhir-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
```diff
@@ -103,9 +102,16 @@ class ExecutionService(jobRepository: IJobRepository, mappingRepository: IMappin
     }
 
     // create execution
-    val mappingJobExecution = FhirMappingJobExecution(executionId.getOrElse(UUID.randomUUID().toString), job = mappingJob, projectId = projectId, mappingTasks = mappingTasks,
-      mappingErrorHandling = executeJobTask.flatMap(_.mappingErrorHandling).getOrElse(mappingJob.dataProcessingSettings.mappingErrorHandling))
-    val fhirMappingJobManager = getFhirMappingJobManager(mappingJob.dataProcessingSettings.mappingErrorHandling)
+    val mappingJobExecution = FhirMappingJobExecution(executionId.getOrElse(UUID.randomUUID().toString), job = mappingJob,
+      projectId = projectId, mappingTasks = mappingTasks)
+    val fhirMappingJobManager = new FhirMappingJobManager(
```
dogukan10, line 107: Let's make it a ExecutionService field so that we can utilize it in both functions.

dogukan10 (changes requested):

tofhir-engine/src/main/scala/io/tofhir/engine/mapping/FhirMappingJobManager.scala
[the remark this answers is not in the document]
dogukan10: OK, if @YemreGurses agrees that it will not break anything, we can resolve this thread.

tofhir-server/src/main/scala/io/tofhir/server/service/ExecutionService.scala
Okanmercan99: When I make it an ExecutionService field, It causes a null pointer error in FhirMappingJobManager#62. It is thrown inside of a library function, and I did not find any reason for this behavior.

dogukan10 (approved):

PR #143, opened 2023-12-15
:sparkles: add support for tsv (tab separated version of csv) files

YemreGurses:


tofhir-engine/src/test/scala/io/tofhir/test/FhirMappingJobManagerTest.scala
```diff
@@ -69,6 +69,11 @@ class FhirMappingJobManagerTest extends AsyncFlatSpec with BeforeAndAfterAll wit
     sourceContext = Map("source" -> FileSystemSource(path = "patients-extra.csv"))
   )
 
+  val patientTsvFileMappingTask: FhirMappingTask = FhirMappingTask(
+    mappingRef = "https://aiccelerate.eu/fhir/mappings/patient-mapping",
+    sourceContext = Map("source" -> FileSystemSource(path = "patients.tsv", options = Map("sep" -> "\\t")))
```
dogukan10, line 74: We need to pass **"sep" -> "\\t"** option to parse a tsv file right ? Maybe, this option can be automatically added in FileDataSourceReader for tsv files, WDYT ?

YemreGurses: I'm not sure. Although the file itself is a tsv file, different character could be used as a separator. So \t seperator is not mandatory. Just as \t can be used in csv files, comma can be used in tsv files.

dogukan10: I think that if you are using another separator then you should change the file format accordingly like CSV (comma separated) or SSV (semicolon-separated). In terms of user experience and consistency, we should handle it automatically.

YemreGurses: What if the user has no permission to change it. Is it better to simply give them an option to use whatever seperator they want?

dogukan10: As a user, if I want to read from a tsv file, then I would not want to specify a separator option because it should be tab by definition. This is a really bad user experience in my opinion.

Moroever, the user can always provide another separator to override the default one.

sinaci: I agree with @dogukan10. Will you be updating accoringly @YemreGurses ?

tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
```diff
@@ -35,15 +35,21 @@ class FileDataSourceReader(spark: SparkSession) extends BaseDataSourceReader[Fil
...
+        case SourceFileFormats.CSV | SourceFileFormats.TSV =>
           //Options that we infer for csv
           val inferSchema = schema.isEmpty || mappingSource.preprocessSql.isDefined
           val csvSchema = if(mappingSource.preprocessSql.isDefined) None else schema
           //val enforceSchema = schema.isDefined
           val includeHeader = mappingSource.options.get("header").forall(_ == "true")
-          val fileFilter = mappingSource.options.getOrElse("pathGlobFilter", "*.csv")
-          //Other options
-          val otherOptions = mappingSource.options.filterNot(o => o._1 == "header" || o._1 == "inferSchema" || o._1 == "enforceSchema")
+          val fileFilter = mappingSource.options.getOrElse("pathGlobFilter", s"*.{${SourceFileFormats.CSV},${SourceFileFormats.TSV}}")
```
dogukan10, line 44: This filter might lead some unexpected results. For example, I have a directory which has two CSV and two TSV files. I want to read only TSV files but it will read all of them because of this filter, right ?

YemreGurses: I thought the similar thing but we are setting data source with its extension (patients.csv). So it should not be problem for batch jobs. Regarding streaming ones, could be a problem. I think it's better to refactor here.

tofhir-engine/src/test/scala/io/tofhir/test/FhirMappingJobManagerTest.scala
dogukan10: It's done, thanks @YemreGurses. Closing the thread.

dogukan10 (approved):

tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
```diff
@@ -35,18 +35,29 @@ class FileDataSourceReader(spark: SparkSession) extends BaseDataSourceReader[Fil
...
+                // use *.csv as pathGlobFilter by default if it is not set explicitly to ignore files without csv extension
+                ("pathGlobFilter" -> mappingSource.options.getOrElse("pathGlobFilter", s"*.${SourceFileFormats.CSV}"))
+          }
+
           //Options that we infer for csv
           val inferSchema = schema.isEmpty || mappingSource.preprocessSql.isDefined
           val csvSchema = if(mappingSource.preprocessSql.isDefined) None else schema
           //val enforceSchema = schema.isDefined
           val includeHeader = mappingSource.options.get("header").forall(_ == "true")
-          val fileFilter = mappingSource.options.getOrElse("pathGlobFilter", "*.csv")
```
sinaci, line 44: Does this mean that we are now processing all files under a given folder without any extension filter? It is going to be the user's responsibility to put the files in the correct format w.r.t his/her configuration, am I right?

What will happen if the reader encounters a txt file with some dummy text in it?

YemreGurses: Actually no. Previous limitations remain. We just moved pathGlobFilter param to otherOptions map. And used it in spark.read section. Giving it to the spark explicitly (first one) would be same with the second one if we put pathGlobFilter param in otherOptions.

```
   spark.readStream
          .option("pathGlobFilter",fileFilter) // Ignore files without csv extension in default
          .option("enforceSchema", false) //Enforce schema should be false (See https://spark.apache.org/docs/latest/sql-data-sources-csv.html)
          .option("header", includeHeader)
          .option("inferSchema", inferSchema)
          .options(otherOptions)
          .schema(csvSchema.orNull)
          .csv(finalPath)
```

   ```
     spark.readStream
          .option("enforceSchema", false) //Enforce schema should be false (See https://spark.apache.org/docs/latest/sql-data-sources-csv.html)
          .option("header", includeHeader)
          .option("inferSchema", inferSchema)
          .options(otherOptions)
          .schema(csvSchema.orNull)
          .csv(finalPath)
```

PR #145, opened 2023-12-20
Add custom error message for invalid or missing file format

dogukan10:
Defines a custom exception named MissingFileFormatException and improves FileSystemSource.sourceType function to make use of this custom exception.

### DIFF
--- a/tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingTask.scala
+++ b/tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingTask.scala
@@ -1,6 +1,19 @@
 package io.tofhir.engine.model
 
 /**
+ * Custom exception class for signaling errors related to missing or invalid file formats.
+ *
+ * This exception is thrown when a required file format is not provided or is determined
+ * to be invalid during processing.
+ *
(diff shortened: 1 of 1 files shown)


tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingTask.scala
```diff
@@ -25,12 +38,36 @@ trait FhirMappingSourceContext extends Serializable {
...
+    // the determined format
+    var format = fileFormat
+    // split the path into segments based on the period ('.')
+    val pathSegments = path.split('.')
+
+    // if the file format is empty and there are exactly two segments in the path,
+    // set the file format based on the last segment of the path
+    if (format.isEmpty && pathSegments.length == 2) {
+      format = Some(pathSegments.last)
+    }
```
YemreGurses, line 62: I think when the file name is patients.2.csv, this breaks because of `pathSegments.length == 2` check. Maybe we should split if path has "." char then get the last value.

dogukan10: Updated the if statement accordingly.

PR #151, opened 2023-12-29
Stop batch mapping task and some refactoring

dogukan10:
- Throw an exception if the file path is provided for a streaming job instead of a directory path.
- Refactor some classes by removing the redundant usage of RunningJobRegistry
- Add STOPPED and SKIPPED mapping job statuses
- Remove unused FhirMappingInvalidResourceException class


tofhir-engine/src/main/scala/io/tofhir/engine/execution/RunningJobRegistry.scala
```diff
@@ -35,6 +43,8 @@ class RunningJobRegistry(spark: SparkSession) {
    */
   def registerStreamingQuery(execution: FhirMappingJobExecution, mappingUrl: String, streamingQueryFuture: Future[StreamingQuery], blocking: Boolean = false): Future[Unit] = {
     Future {
+      // If there is an error in the streaming query, stop the mapping execution.
+      streamingQueryFuture.recover(_ => stopMappingExecution(execution.job.id, execution.id, mappingUrl))
```
sinaci, line 47: Why is this recovered here? Is it because to conveniently call stopMappingExecution function of the registry?

tofhir-engine/src/main/scala/io/tofhir/engine/mapping/FhirMappingJobManager.scala
```diff
@@ -270,10 +266,10 @@ class FhirMappingJobManager(
                                        identityServiceSettings: Option[IdentityServiceSettings] = None,
                                        timeRange: Option[(LocalDateTime, LocalDateTime)] = None,
                                        executionId: Option[String] = None
-                                      ): Future[Dataset[FhirMappingResult]] = {
+                                      ): Future[Dataset[FhirMappingResult]] = Future {
```
sinaci, line 269: Why not the following?

= {
    val (fhirMapping, mds, df) = readJoinSourceData(task, sourceSettings, timeRange, jobId = Some(jobId))
    executeTask(jobId, fhirMapping, df, mds, terminologyServiceSettings, identityServiceSettings, executionId)
  }

dogukan10: executeTask returns a Future but readJoinSourceData does not and this prevent us from accessing the errors encountered in readJoinSourceData. Therefore, the logging was not working as expected. By converting it to a Future, I managed to catch errors and handle the logging properly.

tofhir-engine/src/main/scala/io/tofhir/engine/execution/RunningJobRegistry.scala
dogukan10: Let's assume there is an error such as invalid folder path while running the streaming job which actually cancels the job i.e. the mapping job is not running. However, since this error is unknown to the registry, it assumes that it is still running, therefore we can not rerun it. I made this update to stop mapping job in the registry so that we can rerun it.

tofhir-engine/src/main/scala/io/tofhir/engine/mapping/FhirMappingJobManager.scala
sinaci: I see, thanks for the explanation.

I recommend to explain this also as a comment alongside the code.

If sth. like the following also OK, I would recommed the restructure accordingly:

    Future.apply(readJoinSourceData(task, sourceSettings, timeRange, jobId = Some(jobId))) flatMap {
      case (fhirMapping, mds, df) => executeTask(jobId, fhirMapping, df, mds, terminologyServiceSettings, identityServiceSettings, executionId)
    }

PR #162, opened 2024-02-06
Fhir r5 update

YemreGurses:


tofhir-engine/src/main/resources/application.conf
```diff
@@ -17,6 +17,7 @@ tofhir {
       repository = { # The repository where the schema definitions are kept.
         folder-path = "schemas"
       }
+      fhir-version = "4.0.1"
```
dogukan10, line 20: Could you please add a comment here ?

tofhir-server/src/main/scala/io/tofhir/server/fhir/FhirDefinitionsConfig.scala
```diff
@@ -7,6 +7,8 @@ import scala.util.Try
 
 class FhirDefinitionsConfig(fhirDefinitionsConfig: Config) {
 
+  /** Major FHIR version (R4 or R5) */
+  lazy val majorFhirVersion: String = Try(fhirDefinitionsConfig.getString("fhir-version")).getOrElse("R5")
```
dogukan10, line 11: If we are planning to use R5 by default, let's change other files accordingly including the fhir-version settings in application.conf file. If the default version is R4, let's change it **getOrElse("R4")**

tofhir-server/src/main/resources/application.conf
```diff
@@ -17,6 +17,7 @@ tofhir {
       repository = { # The repository where the schema definitions are kept.
         folder-path = "schemas"
       }
+      fhir-version = "4.0.1"
```
dogukan10, line 20: Let's add a comment here.

dogukan10 (approved):

PR #165, opened 2024-02-13
Implement an endpoint to proxy requests to tofhir-redcap

dogukan10:
This merge request adds a new endpoint, enabling the proxying of requests to the tofhir-redcap microservice.


sinaci:
I was expecting to retrieve the REDCap connection/configuration parameters from the backend and then web client to continue the communication with REDCap as it was. I see that all REDCap communication is now proxied through the backend. This looks better 👍🏼

PR #172, opened 2024-03-15
Update spark version. Add cluster configs to ToFhirConfig. Fix sink p…

Okanmercan99:
…ath bug for hadoop.

### DIFF
--- a/tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
+++ b/tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
@@ -8,6 +8,8 @@ import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{DataFrame, SparkSession}
 
 import java.io.File
+import java.net.URI
+import java.nio.file.Paths
 import java.time.LocalDateTime
 import scala.collection.mutable
 
@@ -32,9 +34,9 @@ class FileDataSourceReader(spark: SparkSession) extends BaseDataSourceReader[Fil
    */
   override def read(mappingSource: FileSystemSource, sourceSettings:FileSystemSourceSettings, schema: Option[StructType], timeRange: Option[(LocalDateTime, LocalDateTime)], limit: Option[Int] = Option.empty,jobId: Option[String] = Option.empty): DataFrame = {
 
-    // Do not add absolute path if it is a hadoop path
+    // Do not add context path if it is a hadoop path
     val finalPath = if (sourceSettings.dataFolderPath.startsWith("hdfs://")) {
-      sourceSettings.dataFolderPath + "/" + mappingSource.path
+      new URI(s"${sourceSettings.dataFolderPath.stripSuffix("/")}/${mappingSource.path.stripPrefix("/")}").toString
     } else {
       FileUtils.getPath(sourceSettings.dataFolderPath, mappingSource.path).toAbsolutePath.toString
     }
--- a/tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
+++ b/tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
@@ -31,7 +31,15 @@ class FileDataSourceReader(spark: SparkSession) extends BaseDataSourceReader[Fil
    * @throws NotImplementedError      If the specified source format is not implemented.
    */
   override def read(mappingSource: FileSystemSource, sourceSettings:FileSystemSourceSettings, schema: Option[StructType], timeRange: Option[(LocalDateTime, LocalDateTime)], limit: Option[Int] = Option.empty,jobId: Option[String] = Option.empty): DataFrame = {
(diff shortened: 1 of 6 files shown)


tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
```diff
@@ -31,7 +31,15 @@ class FileDataSourceReader(spark: SparkSession) extends BaseDataSourceReader[Fil
    * @throws NotImplementedError      If the specified source format is not implemented.
    */
   override def read(mappingSource: FileSystemSource, sourceSettings:FileSystemSourceSettings, schema: Option[StructType], timeRange: Option[(LocalDateTime, LocalDateTime)], limit: Option[Int] = Option.empty,jobId: Option[String] = Option.empty): DataFrame = {
-    val finalPath = FileUtils.getPath(sourceSettings.dataFolderPath, mappingSource.path).toAbsolutePath.toString
+
+    // Do not add absolute path if it is a hadoop path
```
sinaci, line 35: Do you mean the context path?

tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
```diff
@@ -31,7 +31,15 @@ class FileDataSourceReader(spark: SparkSession) extends BaseDataSourceReader[Fil
...
+    val finalPath = if (sourceSettings.dataFolderPath.startsWith("hdfs://")) {
+      sourceSettings.dataFolderPath + "/" + mappingSource.path
```
sinaci, line 37: Please use Paths to generate new paths.

tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
```diff
@@ -31,7 +31,15 @@ class FileDataSourceReader(spark: SparkSession) extends BaseDataSourceReader[Fil
    * @throws NotImplementedError      If the specified source format is not implemented.
    */
   override def read(mappingSource: FileSystemSource, sourceSettings:FileSystemSourceSettings, schema: Option[StructType], timeRange: Option[(LocalDateTime, LocalDateTime)], limit: Option[Int] = Option.empty,jobId: Option[String] = Option.empty): DataFrame = {
-    val finalPath = FileUtils.getPath(sourceSettings.dataFolderPath, mappingSource.path).toAbsolutePath.toString
+
+    // Do not add absolute path if it is a hadoop path
+    val finalPath = if (sourceSettings.dataFolderPath.startsWith("hdfs://")) {
```
sinaci, line 36: Does HDFS have other attributes which may be included as config parameters in DataSourceSettings? If so, it will be better to implement a dedicated class for HDFS and implement its specific reader instead of this "if" statement.

tofhir-rxnorm/src/main/scala/io/tofhir/rxnorm/PullRxNormNdcMedDetails.scala
```diff
@@ -0,0 +1,143 @@
+package io.tofhir.rxnorm
+
+import java.io.{FileReader, FileWriter}
+import com.opencsv.{CSVReaderBuilder, CSVWriter}
+import io.onfhir.api.util.FHIRUtil
+import org.json4s.JsonAST.{JArray, JDouble, JObject, JString}
+import org.json4s._
+import io.onfhir.util.JsonFormatter._
+
+object PullRxNormNdcMedDetails extends App {
```
sinaci, line 10: What does this class do? Please add a detailed comment to describe. Please give input and output examples.

tofhir-rxnorm/src/main/scala/io/tofhir/rxnorm/PullRxNormNdcMedDetails.scala
```diff
@@ -0,0 +1,143 @@
...
+  // RxNorm Api client for calls
+  val rxNormApiClient = RxNormApiClient("https://rxnav.nlm.nih.gov", 10)
+
+  // Read ncd codes from prescription table
+  val csvFilePath = "mimic-iv-data/hosp/prescriptions.csv"
```
sinaci, line 15: How do these variables used? Please describe with examples.

sinaci (changes requested):

tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
Okanmercan99: I refactored using the URI class for this case because the path contains a ":" character, which is illegal for Paths.

tofhir-engine/src/main/scala/io/tofhir/engine/data/read/FileDataSourceReader.scala
Okanmercan99: HDFS has no other attributes we can pass using config parameters in DataSourceSettings.

PR #183, opened 2024-05-09
Reduce the memory usage of local terminology service

dogukan10:
- Enhance CsvUtil.readFromCSV to read rows iteratively from CSV file, minimizing memory usage. Results in an average reduction of 0.1 GB in memory consumption.
- Trigger Garbage Collection in LocalTerminologyService immediately after populating concept maps and code systems to reduce memory usage by 0.5 GB on average.
- Eliminate "lazy" keyword from FhirMappingService.terminologyService variable to ensure creation of Local Terminology Service only once, preventing multiple creations caused by the "lazy" keyword.

### DIFF
--- a/docker/server/docker-entrypoint.sh
+++ b/docker/server/docker-entrypoint.sh
@@ -1,6 +1,13 @@
 #!/usr/bin/env bash
 
-JAVA_CMD="java -Xms256m -Xmx3g -jar "
+# Set default Java options
+DEFAULT_JAVA_OPTIONS="-Xms256m -Xmx3g"
+
+# Use environment variable if provided, otherwise use default value
+JAVA_OPTIONS="${JAVA_OPTIONS:-$DEFAULT_JAVA_OPTIONS}"
+
+# Construct JAVA_CMD with Java options
+JAVA_CMD="java $JAVA_OPTIONS -jar "
 
 # Configure application.conf path
 if [ ! -z "$APP_CONF_FILE" ]; then


YemreGurses:
I have tried running different mappings with this branch and it may not be better than previous implementation. Results of running with -Xms256m -Xmx5g and a single mapping task using a CSV file with 500.000 rows as a data source:
Mappings were successfully done on the main branch, but failed due to memory error on this implementation. 

Also, the duration of regular mappings seems to have increased a lot. Even a mapping job that generated only 3 FHIR resources took too long to complete (~3 minutes 45 seconds). Same setup is completed in ~20 seconds at the main branch.

dogukan10:
> I have tried running different mappings with this branch and it may not be better than previous implementation. Results of running with -Xms256m -Xmx5g and a single mapping task using a CSV file with 500.000 rows as a data source: Mappings were successfully done on the main branch, but failed due to memory error on this implementation.
> 
> Also, the duration of regular mappings seems to have increased a lot. Even a mapping job that generated only 3 FHIR resources took too long to complete (~3 minutes 45 seconds). Same setup is completed in ~20 seconds at the main branch.

I anticipate that mappings may take longer due to the recent adjustment in CsvUtil.readFromCSV, where instead of loading the entire file content at once, it now reads it line by line. I've reversed this modification. Could you please rerun your tests and inform me whether mappings are successfully completed in this branch as well, and if the duration of mappings is OK? If not, I'll also undo the commit associated with the Garbage Collector since the main problem is resolved by JAVA_OPTIONS environment variable now.

PR #185, opened 2024-05-22
Build Docker images for Fluentd and Kibana; Address Kibana-related bugs

dogukan10:


tofhir-engine/src/main/scala/io/tofhir/engine/mapping/MappingTaskExecutor.scala
```diff
@@ -164,7 +165,8 @@ object MappingTaskExecutor {
                   code = FhirMappingErrorCodes.INVALID_INPUT,
                   description = validationErrors.mkString("\n")
                 )),
-                executionId = executionId
+                executionId = executionId,
+                projectId = fhirMappingService.projectId
```
sinaci, line 169: Does this create a kind of dependency from tofhir-engine to the project concept that we use in tofhir-server? If yes, does tofhir-engine still work by itself as a runnable jar?

dogukan10: The optional **projectId** field in the **FhirMappingResult** model does not affect the standalone functionality of tofhir-engine. The projectId field is only utilized within tofhir-server where the project concept is relevant.

PR #191, opened 2024-06-10
:sparkles: add metadata endpoint to retrieve version & config informa…

YemreGurses:
…tion through API


tofhir-server/src/main/scala/io/tofhir/server/model/Metadata.scala
```diff
@@ -0,0 +1,31 @@
+package io.tofhir.server.model
+
+/**
+ * Model that represents the metadata of the server.
+ */
+case class Metadata(name: String,
```
dogukan10, line 6: It would be good to add parameter documentations as well.

tofhir-server/src/main/scala/io/tofhir/server/service/MetadataService.scala
```diff
@@ -0,0 +1,88 @@
...
+    val proxiedRequest = HttpRequest(
+      method = HttpMethods.GET,
+      uri = s"${redCapServiceConfig.endpoint}/$SEGMENT_METADATA",
+      headers = RawHeader("Content-Type", "application/json") :: Nil
+    )
+
+    val responseFuture: Future[HttpResponse] = Http().singleRequest(proxiedRequest)
+    val responseAsString = Try(Await.result(
+      responseFuture.flatMap(resp => Unmarshal(resp.entity).to[String]),
+      1.seconds
```
dogukan10, line 78: We can increase it to 5-10 seconds maybe ?

dogukan10 (changes requested):

dogukan10 (approved):

PR #211, opened 2024-08-12
Implement base FHIR profile retrieval and CSV writing in addition to the improved documentation of handleJoin

dogukan10:
- Implemented an endpoint to retrieve base FHIR profiles
- Modified `FileSystemWriter` to write generated resources to a CSV file
- Modified `FhirDefinitionsService` to return simplified element definitions of a schema
- Refactored `handleJoin` method and improved its documentation
- Added a test case for joining multiple FHIR resources and writing to a CSV file

### DIFF
--- a/tofhir-server/src/main/scala/io/tofhir/server/endpoint/SchemaDefinitionEndpoint.scala
+++ b/tofhir-server/src/main/scala/io/tofhir/server/endpoint/SchemaDefinitionEndpoint.scala
@@ -33,7 +33,7 @@ class SchemaDefinitionEndpoint(schemaRepository: ISchemaRepository, mappingRepos
       pathEndOrSingleSlash {
         parameterMap { queryParams =>
           queryParams.get("url") match {
-            case Some(url) => getSchemaByUrl(projectId, url)
+            case Some(url) => getSchemaByUrl(url)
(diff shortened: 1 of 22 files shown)


tofhir-server/src/main/scala/io/tofhir/server/endpoint/SchemaDefinitionEndpoint.scala
```diff
@@ -114,10 +114,10 @@ class SchemaDefinitionEndpoint(schemaRepository: ISchemaRepository, mappingRepos
     }
   }
 
-  private def getSchemaByUrl(projectId: String, url: String): Route = {
+  private def getSchemaByUrl(url: String): Route = {
     get {
       complete {
-        service.getSchemaByUrl(projectId, url) map {
+        service.getSchemaByUrl(url) map {
```
YemreGurses, line 120: I guess this is because schema urls are already unique reagardless of project.

dogukan10: Yes, that's right.

YemreGurses:
It not urgent but we may update api.yaml file for new endpoints maybe after merging.

YemreGurses (approved):

PR #218, opened 2024-08-29
Prefix management

sinaci:


YemreGurses (commented):
We should handle an empty schema folder case (e.g. schemas/icrc) and throw an exception accordingly.

sinaci:
> We should handle an empty schema folder case (e.g. schemas/icrc) and throw an exception accordingly.

It is now handled with b442e65a186f28d832cc171c79178852be7f2f26

PR #220, opened 2024-09-05
:sparkles: feat(FileDataSourceReader): Add txt-tsv reading.

Okanmercan99:


sinaci:
We do not need this PR anymore. We decided to use an option in sourceBinding indicating the content-type of the file being read whether it comes from a zip file or from a folder.

PR #221, opened 2024-09-05
Enhance FileSystemWriter: Add Support for Parquet and Delta Lake, Advanced Partitioning, and Comprehensive Tests

dogukan10:
- **Enhance FileSystemWriter**: Update the FileSystemWriter to support writing to Parquet and Delta Lake formats.
- **Implement Partitioning**: Enable partitioning by resource type as well as additional columns in the DataFrame.
- **Add FileSystemWriter Tests**: Implement comprehensive test cases to validate the functionality of FileSystemWriter.


tofhir-engine/src/main/scala/io/tofhir/engine/data/write/FileSystemWriter.scala
```diff
@@ -18,15 +21,129 @@ class FileSystemWriter(sinkSettings: FileSystemSinkSettings) extends BaseFhirWri
...
+        // +------------+-------------------------+
+        // | resourceType | mappedResourceJson    |
+        // +------------+-------------------------+
+        // | Patient      | {"resourceType": "... |
+        // | Patient      | {"resourceType": "... |
+        // +------------+-------------------------+
+        val parsedDF = df
+          .map(row => {
+            // extract `mappedResource` and parse the resourceType from the JSON.
+            val mappedResourceJson = row.mappedResource.get
```
sinaci, line 38: We parse the JSON string in the dataframe to find the resource type at this point. What if we put the resource type into FhirMappingResult and set it before we serialize the JObject to String within MappingTaskExecutor? If we do that, I believe we can remove this parsing and improve performance.

What do you think?

dogukan10: I'll extend the `FhirMappingResult` model by adding an optional `resourceType` field. It will significantly improve the performance, as we can use this field directly to group the DataFrame.

dogukan10: Done https://github.com/srdc/tofhir/commit/63cb8a99d2a1cf33ca76358739c099f491428907

PR #226, opened 2024-09-18
Refactor file format to content type

Okanmercan99:


Okanmercan99:
image
For this update, you should change the `fileFormat` field to `contentType`. Additionally, you should add the `contentType` field to all `FileSystemSource` entries.

Okanmercan99:
image
You should add the `contentType` field to all `sinkSettings` that are `FileSystemSinkSettings`.

dogukan10:
I believe that `contentType` should be optional. In most cases, such as the example below, specifying `contentType` is redundant because the file extension (e.g., `.csv`) already indicates the content type:

```json
"sourceBinding": {
  "source": {
    "jsonClass": "FileSystemSource",
    "path": "patients.csv",
    "contentType": "csv",
    "sourceRef": "pilot1-source"
  }
}
```

The use of `contentType` should be limited to scenarios where a file extension can represent multiple content types. For instance, a `.txt` file could contain either CSV or NDJSON data, so in that case, `contentType` would clarify the format (e.g., `csv` or `ndjson`) while reading the file.

WDYT @YemreGurses, shall we make it optional ?

dogukan10 (changes requested):

dogukan10 (approved):

Issue #227, opened 2024-09-19
Bug with /file-system-path endpoint


YemreGurses:
This URL is working when the server is running locally: 
`http://localhost:8085/tofhir/file-system-path`
`http://localhost:8085/tofhir/file-system-path?base_path=/&include_files=false`

However, after deployed with docker, GET requests to these endpoints does not give any response and times out e.g. 
`https://matrix.srdc.com.tr/dt4h/tofhir/api/file-system-path`
`https://matrix.srdc.com.tr/dt4h/tofhir/api/file-system-path?base_path=/&include_files=false`

PR #231, opened 2024-09-26
fix FileSystemTreeStructureService bug

YemreGurses:
fix FileSystemTreeStructureService to work with unix paths and restrict paths from going outside configured context path

fixes: https://github.com/srdc/tofhir/issues/227

### DIFF
--- a/tofhir-server/src/main/scala/io/tofhir/server/service/fhir/FileSystemTreeStructureService.scala
+++ b/tofhir-server/src/main/scala/io/tofhir/server/service/fhir/FileSystemTreeStructureService.scala
@@ -2,6 +2,7 @@ package io.tofhir.server.service.fhir
 
 import io.tofhir.engine.util.FileUtils
 import io.tofhir.server.model.FilePathNode
+import io.tofhir.server.service.fhir.FileSystemTreeStructureService.FILE_LIMIT
 
 import java.nio.file.{Files, Path}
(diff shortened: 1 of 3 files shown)


tofhir-server/src/main/scala/io/tofhir/server/service/fhir/FileSystemTreeStructureService.scala
```diff
@@ -25,20 +26,42 @@ class FileSystemTreeStructureService {
...
+    if(!path.toAbsolutePath.normalize().startsWith(FileUtils.getPath("").toAbsolutePath)) {
+      throw new IllegalArgumentException("The given path is outside the root context path.")
+    }
     buildNode(path)
   }
 
+}
+
+object FileSystemTreeStructureService {
+  val FILE_LIMIT = 10000 // Maximum number of files to process
```
dogukan10, line 66: Should we make it configurable via `application.conf` ?

YemreGurses: Sure, but I'm not sure where should we put this config. It is a tofhir-server specific config and we currently don't have a config for tofhir-server specifically. There is web-server config but it's more related to http stuff. For now, let's not define a separate config file just for this. It would be better if it was configurable, but there is no need for it to be configurable at the moment. If we need tohir-server specific configs in the future, we can add that one as well.

PR #232, opened 2024-09-26
:white_check_mark: try fixing the test: run a batch mapping job which…

YemreGurses:
… converts FHIR Resources into a flat schema and write them to a CSV file

### DIFF
--- a/tofhir-server/src/test/scala/io/tofhir/server/endpoint/MappingExecutionEndpointTest.scala
+++ b/tofhir-server/src/test/scala/io/tofhir/server/endpoint/MappingExecutionEndpointTest.scala
@@ -377,24 +377,29 @@ class MappingExecutionEndpointTest extends BaseEndpointTest with OnFhirTestConta
         // Mappings run asynchronously. Wait at most 45 seconds for mappings to complete.
         val success = waitForCondition(45) {
           fsSinkFolder.listFiles.exists(_.getName.contains("results.csv")) && {
-            // read the csv file created in the file system
-            val outputFolder: File = fsSinkFolder.listFiles.find(_.getName.contains("results.csv")).get
-            val results = sparkSession.read
-              .option("header","true")
-              .csv(outputFolder.getPath)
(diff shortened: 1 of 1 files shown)


tofhir-server/src/test/scala/io/tofhir/server/endpoint/MappingExecutionEndpointTest.scala
```diff
@@ -377,24 +377,31 @@ class MappingExecutionEndpointTest extends BaseEndpointTest with OnFhirTestConta
...
+              results.count() == 1
+            }
+            catch {
+              case e: Exception => false
+            }
           }
         }
+        if (success) {
+          // read the csv file created in the file system
+          val outputFolder: File = fsSinkFolder.listFiles.find(_.getName.contains("results.csv")).get
```
sinaci, line 392: Why do we read here again? Cannot we use the results from above?

tofhir-server/src/test/scala/io/tofhir/server/endpoint/MappingExecutionEndpointTest.scala
```diff
@@ -377,24 +377,31 @@ class MappingExecutionEndpointTest extends BaseEndpointTest with OnFhirTestConta
...
-            // verify the content of row
-            val row = results.head()
-            row.getAs[String]("active").contentEquals("true")
-            row.getAs[String]("birthDate").contentEquals("1974-12-25")
-            row.getAs[String]("gender").contentEquals("male")
-            row.getAs[String]("homeCountry").contentEquals("Australia")
-            row.getAs[String]("id").contentEquals("db03a265-0194-460e-8625-69e25daec2d7")
-            row.getAs[String]("officialName").contentEquals("Peter Chalmers")
-            row.getAs[String]("phone").contentEquals("(03) 5555 6473")
+            try {
```
dogukan10, line 380: Let's add a comment explaining the necessity of the try/catch block here.

PR #238, opened 2024-10-04
Replace hardcoded inputType and returnType fields of FhirPathFunctions with FHIR_DATA_TYPEs values

camemre49:


dogukan10:
Please note that this MR does not work. When I call 
http://localhost:8085/tofhir/fhir-path-functions, it returns empty input and return types. I think the following function should be updated to handle enumerations: https://github.com/srdc/onfhir/blob/master/onfhir-path/src/main/scala/io/onfhir/path/AbstractFhirPathFunctionLibrary.scala#L44. Marking it as draft.

PR #242, opened 2024-10-11
Improve fhir mapping result

YemreGurses:


tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingResult.scala
```diff
@@ -5,44 +5,75 @@ import io.tofhir.engine.util.TimeUtil
...
+
+  /**
+   * The optional identifier for the project under which this mapping job is being executed.
+   */
+  val projectId: Option[String]
+}
 
 /**
- * Mapping process result
+ * Case class representing the result of a FHIR mapping process, extending the base mapping result.
```
sinaci, line 50: Can you please extend the description for this class?

FhirMappingResult class represents a single mappedFhirResource irrespective of how many FHIR resources is created from each row. Hence, different FhirMappingResult instances might be created from a single source row if the mapping is 1-to-many.

tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingResult.scala
```diff
@@ -5,44 +5,75 @@ import io.tofhir.engine.util.TimeUtil
...
+
+  /**
+   * Timestamp of the result
+   */
+  val timestamp: Timestamp
+
+  /**
+   * If there is a problem in the process, the JSON serialization of the source data
+   */
+  val source: String
```
sinaci, line 31: Is the "source" field always populated? Then, I suggest the update the description accordingly.

tofhir-engine/src/main/scala/io/tofhir/engine/model/FhirMappingResult.scala
```diff
@@ -73,20 +104,41 @@ case class FhirMappingResult(
...
+ * Stores the results of mapping i.e. generated FHIR resources for a source input.
+ * It groups the mapped FHIR resources generated from the input data, allowing for easy access to
+ * the output for a specific input row.
+ *
+ * In addition to the common fields from [[BaseFhirMappingResult]], it includes a collection of mapped
+ * FHIR resources, which represent the outcome of the mapping process for each input source.
+ *
+ * @param mappedFhirResources List of mapped FHIR resources. This is used to group mapped resources by input row.
+ */
+case class FhirMappingResultsForInput(
```
sinaci, line 123: Can you please extend the description for this class?

FhirMappingResultForInput class represents a the result from a single source row, that is each source row corresponds to one FhirMappingResultForInput instance. If the mapping is 1-to-many, mappedFhirResources field includes all the created FHIR resources from the source row.

sinaci (changes requested):

PR #249, opened 2024-11-15
:sparkles: feat: Implement reload endpoint.

Okanmercan99:


tofhir-server/src/main/scala/io/tofhir/server/repository/mapping/ProjectMappingFolderRepository.scala
```diff
@@ -31,7 +31,7 @@ class ProjectMappingFolderRepository(mappingRepositoryFolderPath: String, projec
   private val logger: Logger = Logger(this.getClass)
 
   // project id -> mapping id -> mapping
-  private val mappingDefinitions: mutable.Map[String, mutable.Map[String, FhirMapping]] = initMap(mappingRepositoryFolderPath)
+  private var mappingDefinitions: mutable.Map[String, mutable.Map[String, FhirMapping]] = initMap(mappingRepositoryFolderPath)
```
YemreGurses, line 34: I'm not sure about this change. Maybe we can continue using `val` and then use [`clear`](https://scala-lang.org/api/current/scala/collection/mutable/Map.html#clear-94c) and `++=` operator. WDYT? @dogukan10

YemreGurses (changes requested):

PR #256, opened 2024-11-25
add timeout error message for mapping test

oburakcinar:


tofhir-server/src/main/scala/io/tofhir/server/endpoint/JobEndpoint.scala
```diff
@@ -42,7 +42,25 @@ class JobEndpoint(jobRepository: IJobRepository, mappingRepository: IMappingRepo
...
+              addCORSHeaders(
+                HttpResponse(
+                  status = StatusCodes.RequestTimeout,
+                  entity = HttpEntity.Strict(
+                    ContentTypes.`application/json`,
+                    ByteString(
+                      """Status Code: 408
+                        |Type: https://tofhir.io/errors/RequestTimeout
+                        |Title: Request Timeout
+                        |Detail: There are too many rows in the data source CSV.
```
YemreGurses, line 55: Let 's update this message: The server could not complete the test because of the large data. Try increasing the timeout config and test again.

tofhir-server/src/main/scala/io/tofhir/server/endpoint/JobEndpoint.scala
```diff
@@ -42,7 +42,25 @@ class JobEndpoint(jobRepository: IJobRepository, mappingRepository: IMappingRepo
...
           pathEndOrSingleSlash {
-            testMappingWithJob(projectId, jobId)
+            withRequestTimeoutResponse { _ =>
+              addCORSHeaders(
+                HttpResponse(
+                  status = StatusCodes.RequestTimeout,
+                  entity = HttpEntity.Strict(
+                    ContentTypes.`application/json`,
+                    ByteString(
+                      """Status Code: 408
```
YemreGurses, line 52: Instead of writing plain string here, can you add a new case class to ToFhirError file for timeout errors. And use that class to construct the error.

YemreGurses (changes requested):
Build failed, one of the imports was wrong, fixed it now FYI.

YemreGurses:
Fixes: https://gitlab.srdc.com.tr/onfhir/tofhir-web/-/issues/96

PR #263, opened 2024-12-04
Handle request timeouts and improve some error messages

camemre49:
Implemented:
 - If request cannot be completed in time, send an appropriate response
 - Validate the mapping test request, if request is not valid send  an appropriate response


YemreGurses (approved):
This PR is ok but we need to show the specific error on the frontend as well. Currently it only says "An unexpected error occurred. Check the developer console for more detail." Can you fix that before merging this PR?

PR #264, opened 2024-12-06
:sparkles: feat: Log kafka topic in the kafka topic not found error.

Okanmercan99:


tofhir-engine/src/main/scala/io/tofhir/engine/execution/RunningJobRegistry.scala
```diff
@@ -97,7 +97,10 @@ class RunningJobRegistry(spark: SparkSession) {
         updatedExecution.getStreamingQuery(mappingTaskName).awaitTermination()
       }catch{
         case exception: StreamingQueryException =>{
-          ExecutionLogger.logExecutionStatus(execution, FhirMappingJobResult.FAILURE, Some(mappingTaskName), Some(exception), isChunkResult = false)
+          val topicNames = execution.mappingTasks.find(mappingTask => mappingTask.name.contentEquals(mappingTaskName)).get.sourceBinding.map(source => source._2.asInstanceOf[KafkaSource].topicName).mkString(", ")
```
dogukan10, line 100: Let's use getCause function to detect org.apache.kafka.common.errors.UnknownTopicOrPartitionException and then add the missing topics to the error message.

The first cause is ExecutionException and its cause is UnknownTopicOrPartitionException. You can add a match statement accordingly.

dogukan10 (changes requested):

tofhir-engine/src/main/scala/io/tofhir/engine/execution/RunningJobRegistry.scala
```diff
@@ -97,7 +98,15 @@ class RunningJobRegistry(spark: SparkSession) {
         updatedExecution.getStreamingQuery(mappingTaskName).awaitTermination()
       }catch{
         case exception: StreamingQueryException =>{
-          ExecutionLogger.logExecutionStatus(execution, FhirMappingJobResult.FAILURE, Some(mappingTaskName), Some(exception), isChunkResult = false)
+          exception.getCause.getCause match{
```
dogukan10, line 101: getCause function can return null and we should handle it.

dogukan10 (approved):

PR #273, opened 2025-01-06
Implement functionality to write data to HDFS

dogukan10:


tofhir-engine/src/main/scala/io/tofhir/engine/data/write/FileSystemWriter.scala
```diff
@@ -22,92 +26,156 @@ class FileSystemWriter(sinkSettings: FileSystemSinkSettings) extends BaseFhirWri
...
+                val conf = new Configuration()
+                conf.set("fs.defaultFS", defaultFS) // Set the HDFS default file system dynamically
+                // Get the HDFS FileSystem instance using the configuration
+                val fs = FileSystem.get(conf)
+
+                // Define the HDFS directory path for the current resourceType
+                val dirPath = new org.apache.hadoop.fs.Path(s"${sinkSettings.path}/$resourceType")
+                // Ensure the directory exists; create it if it does not
+                if (!fs.exists(dirPath)) {
+                  fs.mkdirs(dirPath)
```
sinaci, line 58: I suggest to keep an eye on the exception(s) that may be thrown if the file cannot be created on hadoop because of several reasons such as lack of permission.

sinaci (approved):

PR #278, opened 2025-01-15
:sparkles: feat: Infer required fields.

Okanmercan99:


tofhir-server/src/main/scala/io/tofhir/server/service/SchemaDefinitionService.scala
```diff
@@ -114,7 +115,6 @@ class SchemaDefinitionService(schemaRepository: ISchemaRepository, mappingReposi
     val dataFrame = try {
       SourceHandler
         .readSource(inferTask.name, ToFhirConfig.sparkSession, inferTask.sourceBinding, inferTask.mappingJobSourceSettings.head._2, schema = None)
-        .limit(1) // It is enough to take the first row to infer the schema.
```
YemreGurses, line 117: Understood that this is removed to infer the column is required or not. But, if we make inferring using a large CSV, will we have performance problems because it reads all?

Okanmercan99: How can we determine if a column is mandatory without reviewing all rows? We need to ensure that every row contains a specific value for this column to confirm its necessity.

PR #286, opened 2025-03-11
:sparkles: feat: Improve Infer schema functionality by directly readi…

sinaci:
…ng the jdbc metadata for SQL sources.

### DIFF
--- a/tofhir-engine/src/main/scala/io/tofhir/engine/mapping/schema/SchemaConverter.scala
+++ b/tofhir-engine/src/main/scala/io/tofhir/engine/mapping/schema/SchemaConverter.scala
@@ -9,6 +9,9 @@ import io.onfhir.definitions.common.model.{DataTypeWithProfiles, SimpleStructure
 import io.tofhir.engine.util.MajorFhirVersion
 import org.apache.spark.sql.types._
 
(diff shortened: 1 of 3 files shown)


tofhir-engine/src/main/scala/io/tofhir/engine/mapping/schema/SchemaConverter.scala
```diff
@@ -149,4 +153,96 @@ class SchemaConverter(majorFhirVersion: String) {
...
+        } catch {
+          case e: SQLException =>
+            throw new RuntimeException(s"Failed to establish JDBC connection: ${e.getMessage}", e)
+          case e: Throwable =>
+            throw new RuntimeException(s"Unexpected error while establishing JDBC connection: ${e.getMessage}", e)
+        }
+        if (isQuery) {
+          // For SQL queries, execute the query with a limit of zero rows
+          val statement = use(connection.createStatement())
+          val sqlWithLimit = s"($sqlOrTable) LIMIT 0"
```
Okanmercan99, line 187: What happens if an SQL query already contains a LIMIT clause?

sinaci: OK, I handled it using a regular expression to remove LIMIT caluse(s) if already exists.

Issue #293, opened 2025-07-31
Lab Measurement Conversions


YemreGurses:
In most cases, lab code conversions need to be done together with unit conversions. For this reason, we often define two separate files using the same lab codes—one for concept mapping and another for unit conversion. However, it would be more efficient to manage both conversions in a single CSV file, such as a mapping-context file.

For example, I cannot use the CSV below as a unit conversion file because it includes the `target_code` column. To make it usable, I would need to remove the `target_code`. But when I remove the `target_code` column, I cannot use it as a concept map.

| source_code | source_unit     | target_code | target_unit     | conversion_function     |
|-------------|------------------|-------------|------------------|--------------------------|
| 14682-9     | µmol/L           | 2160-0      | mg/L             | `$this * 0.1131`         |
| 22748-8     | mmol/L           | 22748-8     | mmol/L           | `$this`                  |
| 61152-5     | g/L              | 1751-7      | g/L              | `$this`                  |
| 6598-7      | µg/L             | 6598-7      | ng/L             | `$this * 1000`           |
| 22664-7     | mmol/L           | 14937-7     | mmol/L           | `$this * 0.47`           |
| 62238-1     | ml/min/1.73m2    | 69405-9     | ml/min/1.73m2    | `$this`                  |
| 30934-4     | pmol/L           | 30934-4     | pg/mL            | `$this * 3.46`           |
| 6298-4      | mmol/L           | 2823-3      | mmol/L           | `$this`                  |
| 83107-3     | pg/mL            | 33762-6     | pg/mL            | `$this`                  |
| 14646-4     | mmol/L           | 14646-4     | mmol/L           | `$this`                  |
| 20567-4     | ng/mL            | 2276-4      | ng/mL            | `$this`                  |
| 59260-0     | mmol/L           | 718-7       | g/L              | `$this * 16.114`         |
| 100158-5    | mg/L             | 1754-1      | mg/L             | `$this`                  |
| 14927-8     | mmol/L           | 14927-8     | mmol/L           | `$this`                  |
| 30522-7     | mg/L             | 30522-7     | mg/L             | `$this`                  |
| 41995-2     | mmol/mol         | 41995-2     | mmol/mol         | `$this`                  |
| 14647-2     | mmol/L           | 14647-2     | mmol/L           | `$this`                  |
| 14683-7     | mmol/L           | 2161-8      | mg/L             | `$this * 113.12`         |
[...]

PR #295, opened 2025-08-28
Composite Processing of Concept Maps and Unit Conversions

KeremHmd:
A new class is created to handle Mapping Contexts without explicit categorization, allowing them to be used for any context category as long as they hold the necessary data for that category.

https://github.com/srdc/tofhir/issues/293


tofhir-engine/src/main/scala/io/tofhir/engine/mapping/context/MappingContextLoader.scala
```diff
@@ -89,20 +86,37 @@ class MappingContextLoader extends IMappingContextLoader {
    * @param filePath
    * @return
    */
-  private def readConceptMapContextFromCSV(filePath: String): Future[Map[String, Seq[Map[String, String]]]] = {
+  private def readConceptMapContextFromCSV(filePath: String): Future[ConceptMapContext] = {
```
YemreGurses, line 89: Let's update the comment to reflect latest changes for this method. More examples are better.

YemreGurses (changes requested):

PR #301, opened 2026-07-24
♻️ refactor: rename toFHIR → Ignifyr and split the reactor into Community/Enterprise editions

sinaci:
Consolidated **rename + edition-split** branch — both touch the same files, so they land together.

**Rename (clean break):** packages `io.ignifyr.*`, coordinates `io.ignifyr:ignifyr-*`, HOCON `ignifyr.*`, Docker `srdc/ignifyr-*`, docs; version → `2.0-SNAPSHOT`. Legacy `tofhir` intentionally survives only in the `srdc/tofhir-web` image, the `tofhir-redcap` sibling service, the SwaggerHub link, and git history.

**Edition split:** the reactor is now module-separable into a public Community Edition (`ignifyr-cli` + engine/common/sql/file connectors, Apache-2.0) and a private Enterprise Edition (`ignifyr-server` + fhir-server/kafka/json/delta/streaming/scheduling/redcap/observability/terminology). Everything plugs in through the `IgnifyrExtension` ServiceLoader SPI — the engine never names a connector or format directly. A maven-enforcer `bannedDependencies` gate keeps enterprise-only libs out of community modules. Moving a feature between editions is a one-folder move with zero engine changes.

This is the **module-level** split only. Physical extraction to `ignifyr-enterprise` is Phase 2, not this PR.

**Testing:** `mvn -B verify` green. ⚠️ Requires **JDK 11** — under JDK 21 the Spark reflective path surfaces as a spurious `IllegalAccessError` in `ExtensionRegistrySpec`; unsupported-JDK symptom, not a defect.

**33 commits · 496 files · +18,952 / −11,381**


docker/docker-compose.yml
```diff
@@ -1,29 +1,29 @@
...
     volumes:
-      - './data-integration-suite:/usr/local/tofhir/conf'
-      - './tofhir-docker-logs:/usr/local/tofhir/logs'
-  tofhir-web:
-    image: srdc/tofhir-web:latest
-    container_name: tofhir-web
+      - './data-integration-suite:/usr/local/ignifyr/conf'
+      - './ignifyr-docker-logs:/usr/local/ignifyr/logs'
+  ignifyr-web:
+    image: docker.srdc.com.tr/srdc/tofhir-web:latest
```
Okanmercan99, line 21: Should we change tofhir-web:latest to ignifyr-web:latest?

sinaci: Once this image is pushed with its new name, yes we should change then.

ignifyr-engine/src/main/scala/io/ignifyr/engine/config/IgnifyrConfig.scala
```diff
@@ -52,17 +61,18 @@ object ToFhirConfig {
       .setMaster(sparkMaster)
 
     val sparkConfEntries =
```
Okanmercan99: `spark.sql.extensions` is a comma-separated list in Spark, and `ExtensionRegistry` has explicit
code to respect that — line 98 joins contributions from multiple modules instead of overwriting
them. But that merged value is thrown away here: `++` replaces on key collision, and the user's
`spark { }` block is merged last.

So if a user sets `spark.sql.extensions` for any unrelated reason (Iceberg, Sedona, a custom
optimizer rule), the value goes from

    io.delta.sql.DeltaSparkSessionExtension          (contributed by ignifyr-format-delta)

to

    com.example.MyExtension                          (user's value, Delta silently dropped)

The Delta session extension is then never registered, Delta sink writes fail, and nothing in the
error points back at the `spark { }` block that caused it.

Suggestion: keep the additive rule at this seam too — merge as today, then re-join that one key
from both layers:

    val k = "spark.sql.extensions"
    val merged = sparkConfDefaults ++ contributed ++ userConf
    val joined = Seq(contributed.get(k), userConf.get(k)).flatten
                   .flatMap(_.split(",")).map(_.trim).filter(_.nonEmpty).distinct
    if (joined.isEmpty) merged else merged + (k -> joined.mkString(","))

Note this only applies to the list-valued key. Delta's other contribution,
`spark.sql.catalog.spark_catalog`, is genuinely single-valued, so a user override there is
legitimate and should keep winning as it does today.

ignifyr-engine/src/main/scala/io/ignifyr/engine/spi/ExtensionRegistry.scala
```diff
@@ -0,0 +1,151 @@
...
+      )
+    }
+    entries.headOption.map(_._2)
+  }
+
+  /**
+   * Force every registry to materialize — so a duplicate-registration error surfaces at engine
+   * startup rather than in the middle of a job — and log what was loaded.
+   */
+  def init(): Unit = {
```
Okanmercan99, line 148: The scaladoc promises "Force every registry to materialize — so a duplicate-registration error
surfaces at engine startup rather than in the middle of a job", but `scheduler` is not in the
list, and since these are `lazy val`s whose bodies contain the conflict check, not reading one
means not checking it.

So with two scheduling modules installed by mistake, the engine starts cleanly and looks
healthy. The `IllegalStateException` from `singleCapability` only fires later — at
MappingJobLauncher:91 when a scheduled job is first launched, or on the server at
ExecutionService:133 on the first query about scheduled executions.

Worth noting separately that `streaming` is covered only incidentally. The line that reads it,
IgnifyrEngine:65, exists to decide whether to start the archive timer:

    if (ExtensionRegistry.streaming.isDefined) fileStreamInputArchiver.startStreamingArchiveTask()

Its fail-fast side effect is a coincidence. If that line is ever refactored, moved behind
another condition, or removed, streaming loses its duplicate protection silently — and no test
covers the path, so nothing would catch it.

Making the intent explicit here is a one-liner and removes both problems:

    def init(): Unit = {
      sourceConnectors
      sinkProviders
      terminologyServiceProviders
      identityServiceProviders
      cliCommands
      schemaInferrers
      streaming
      scheduler
      ()
    }

The other two registries are fine as-is and do not need adding: `sourceFailureDescriptors` is a
plain `flatMap` with no uniqueness check so it cannot throw, and `sparkConfContributions` is
already forced while the SparkSession is built at IgnifyrEngine:37, before this call.

ignifyr-runtime-streaming/src/main/scala/io/ignifyr/runtime/streaming/StreamingSinkHandler.scala
```diff
@@ -0,0 +1,47 @@
...
+      mappingTaskName: String
+  ): StreamingQuery = {
+    val datasetWrite = (dataset: Dataset[FhirMappingResult], _: Long) =>
+      try {
+        SinkHandler.writeMappingResult(spark, mappingJobExecution, mappingTaskName, dataset, resourceWriter)
+      } catch {
+        case e: Throwable =>
+          logger.error(
+            s"Streaming chunk resulted in error for project: ${mappingJobExecution.projectId}, job: ${mappingJobExecution.jobId}, execution: ${mappingJobExecution.id}, mappingTask: $mappingTaskName",
+            e.getMessage
```
Okanmercan99: `e.getMessage` is a String, so this resolves to scala-logging's
`error(message: String, args: Any*)` overload rather than `error(message: String, cause: Throwable)`.
The varargs form exists to fill `{}` placeholders, and this message has none (it is fully
s-interpolated), so SLF4J/Logback discards the argument instead of appending it. Logback's usual
rescue — treating a trailing extra argument as the exception — does not apply either, because the
argument is a String, not a Throwable.

Net effect: the log line records that a chunk failed, but nothing about why. No exception type, no
message, no stack trace:

    ERROR StreamingSinkHandler - Streaming chunk resulted in error for project: proj-1,
    job: job-7, execution: exec-3, mappingTask: patient-mapping

That is worse here than almost anywhere else, because this is the deliberate swallow path. The
module guide states the intent plainly — "catches per-micro-batch exceptions and only logs (the
stream survives bad chunks)" — so the exception goes nowhere else: it is not rethrown, the job is
not marked FAILURE, and the user is not notified. This log line is the only record of the failure
and the only diagnostic available afterwards. A streaming job can run for days quietly dropping
chunks and leave behind nothing but "something went wrong".

One-word fix:

    logger.error(s"Streaming chunk resulted in error for project: ...", e)

Pre-existing — `main` has the identical line in `SinkHandler.writeStream`, so this is not a
regression from the extraction. Raising it because the code moves in this PR and the fix is
trivial. Note the sibling path at StreamingJobExecutor.scala:62 gets this right: it logs FAILURE
and rethrows, so there the exception does survive.

ignifyr-engine/src/main/scala/io/ignifyr/engine/config/IgnifyrConfig.scala
sinaci: Yes, this is a valid problem, we can fix as suggested. If there are other list-valued config parameters other than "spark.sql.extensions", we should add them to "val k" as well.

ignifyr-engine/src/main/scala/io/ignifyr/engine/spi/ExtensionRegistry.scala
sinaci: OK, I agree.

Okanmercan99 (approved):

test-flow/check-test-tiers.sh
```diff
@@ -0,0 +1,126 @@
...
+    bad "$(basename "$f") starts a container but declares 'package $p' (short tier!) -- move it to $INTEGRATION_PKG   [$f]"
+  fi
+done < <(grep -rlE "$CONTAINER_RE" --include='*.scala' -- */src/test/scala 2>/dev/null | sort)
+[ "$found_any" -eq 1 ] || bad "no container-backed suites found at all -- the grep or the layout changed; this gate is not actually checking anything"
+
+# ---- 2. no bare scalatest-maven-plugin ---------------------------------------
+log "2. No module leaves <scalatest-maven-plugin> unpinned"
+for pom in */pom.xml; do
+  grep -q 'scalatest-maven-plugin' "$pom" || continue
+  if grep -qE '<wildcardSuites>|<membersOnlySuites>|<suffixes>' "$pom"; then
```
Okanmercan99: Invariant 2 greps the whole pom, but each module has two executions — only the test-phase one must be pinned. If a module's <wildcardSuites> is deleted, the integration-test execution's <membersOnlySuites> still matches the alternation, so this reports PASS while the unpinned test execution runs every suite in the module — including the io.ignifyr.integrationtest ones, i.e. containers in mvn test. That's the exact regression this invariant exists to catch.

All 20 modules declaring the plugin already have exactly one <wildcardSuites>, so requiring it specifically passes as-is:
`  if grep -q '<wildcardSuites>' "$pom"; then`

test-flow/check-editions.sh
```diff
@@ -0,0 +1,173 @@
...
+    else
+      bad "$3 — CLI ran but never reported the missing capability/connector"
+    fi
+  }
+  expect_refusal "$SCRIPT_DIR/config/jobs/streaming-watch-job.json" "MissingCapabilityException\|runtime-streaming" "streaming job -> MissingCapabilityException (streaming)"
+  # Kafka job is a stream, so it's refused at the missing streaming capability before connector
+  # resolution — accept either enterprise-gap error.
+  expect_refusal "$SCRIPT_DIR/config/jobs/kafka-redcap-job.json"   "MissingConnectorException\|connector-kafka\|MissingCapabilityException\|runtime-streaming" "Kafka streaming job -> refused (missing streaming capability or Kafka connector)"
+  # list-plugins must not advertise enterprise plugins on the community jar.
+  lp="$(java "${SPARK_OPENS[@]}" -jar "$CLI_JAR" list-plugins 2>/dev/null)"
```
Okanmercan99: rc isn't checked and stderr is discarded, so if java fails for any reason $lp is empty, the grep matches nothing, and we report "no enterprise plugins" — a real leak would pass the same way. expect_refusal above already classifies this properly; this check should too.
```
  lp="$(java "${SPARK_OPENS[@]}" -jar "$CLI_JAR" list-plugins 2>&1)"
  if ! has "$lp" "Installed Ignifyr extension(s):"; then
    bad "list-plugins produced no usable output on the community jar (JVM failed?)"
  elif echo "$lp" | grep -qi "runtime-streaming\|connector-kafka\|redcap"; then
    bad "list-plugins advertises an enterprise plugin on the community jar"
  else
    ok "list-plugins shows no enterprise plugins on the community jar"
  fi
```

test-flow/check-editions.sh
```diff
@@ -0,0 +1,173 @@
...
+ENTERPRISE_CLASSES=(
+  io/ignifyr/runtime/streaming io/ignifyr/runtime/scheduling
+  io/ignifyr/connector/kafka io/ignifyr/connector/fhirserver
+  io/ignifyr/redcap io/ignifyr/format/delta io/ignifyr/format/json io/ignifyr/observability
+  io/ignifyr/sink/omop
+)
+ENTERPRISE_LIBS=( org/apache/spark/sql/kafka io/delta/ it/sauronsoftware/cron4j net/logstash/logback )
+ENTERPRISE_EXT=(
+  io.ignifyr.runtime.streaming io.ignifyr.runtime.scheduling
+  io.ignifyr.connector.kafka io.ignifyr.connector.fhirserver io.ignifyr.redcap io.ignifyr.sink.omop
```
Okanmercan99, line 41: format-delta does register a top-level IgnifyrExtension (DeltaFormatExtension, for the Delta Spark-session wiring), but it's missing from ENTERPRISE_EXT — so the SPI-manifest section, the one this script calls authoritative, never checks whether it leaked into the community jar. The only section that would is the jar-content one, which is explicitly non-fatal and skipped when no jar lister works. CommunityEditionSeparationSpec covers the Delta sink format, not this extension.
```
  io.ignifyr.connector.kafka io.ignifyr.connector.fhirserver io.ignifyr.redcap io.ignifyr.sink.omop
  io.ignifyr.format.delta
```

ignifyr-runtime-streaming/src/test/scala/io/ignifyr/runtime/streaming/StreamingSinkHandlerTest.scala
```diff
@@ -0,0 +1,98 @@
...
+        }),
+        ArgumentMatchers.any()
+      )
+    ).thenThrow(new Exception())
+
+    // Start streaming
+    val streamingQuery =
+      StreamingSinkHandler.writeStream(sparkSession, execution, df, mockWriter, "someMappingTaskName")
+
+    // Wait for data generation for 5 seconds and then terminate the query
```
KeremHmd: StreamingSinkHandlerTest has a fixed timeout set to 5 seconds. That window turned out to be too tight for Spark streaming startup in my lower-spec environment and caused the test to fail there. With small edits around it, I confirmed the timing varies a lot between runs, 16-42 seconds to reach the batch the assertion needs, so any fixed window is fragile. It should still be noted that: this doesn't pose a high-risk during a CI run and the underlying behavior of the test is actually correct.

Rather than expanding the window, the robust fix can be to wait until the two micro-batches have actually run, with a generous cap so it still fails fast if a second batch never arrives.

Okanmercan99:
Closing without merging — everything on this branch is already on `main`.

**How it landed:** #304 was branched off `ignifyr-rename-whole`, added the
onfhir-libs 4.0.0 migration on top, and was rebase-merged into `main` on
2026-08-14. That carried 67 of this PR's 69 commits over with new SHAs. The
remaining two landed on `main` directly afterwards as `101f5fe4` (test-gate
fixes) and `d04ac889` (StreamingSinkHandlerTest de-flake).

**Verified:** `git cherry main ignifyr-rename-whole` reports no unmerged
commits. The only remaining difference between the two is `main` being *ahead*
by the 4.0.0 migration (18 files) — this branch is behind, not ahead.

**Why it shows conflicts:** the merge base is still `d74d75f9`, which predates
both copies of the work, so git sees the same ~500 files added independently on
each side. That is duplicate history rather than competing edits, and resolving
it would only reproduce a tree `main` already has.

